PR #2, opened 2018-09-28
Update installation tasks to speed up playbook execution

### LINKED ISSUE #4: Restructure any conditionals that specify ansible_facts.lvm
* There are multiple task/block conditionals that reference keys in `ansible_facts.lvm`. That is is problematic when the user doesn't have `lvm2` installed before-hand, and the user runs in the playbook in check mode. If `lvm2` is not installed, then `ansible_facts.lvm` won't be available, and Ansible will terminate the playbook execution.
* If the block in `tasks/vg-default.yml` was run in check mode, then that block will not execute, and sequent tasks that rely on `ansible_facts.lvm` will return fatal status.

timflannagan1:
This would also fix #4 and #7.

### DIFF
--- a/tasks/fs-default.yml
+++ b/tasks/fs-default.yml
@@ -2,23 +2,32 @@
 - name: Stat the final device file
   include_tasks: stat_device.yml
 
-- name: Install xfsprogs for xfs file system type
-  package:
-    name: xfsprogs
-    state: present
-  when: volume.fs_type == "xfs" and not ansible_check_mode
+- block:
+  - name: Install xfsprogs for xfs file system type
+    package:
+      name: xfsprogs
+      state: present
+  - set_fact:
+      storage_packages: "{{ storage_packages + ['xfsprogs'] }}"
+  when: "'xfsprogs' not in storage_packages and volume.fs_type == 'xfs' and not ansible_check_mode"
 
-- name: Install e2fsprogs for ext file system type
-  package:
-    name: e2fsprogs
-    state: present
-  when: volume.fs_type in ["ext2", "ext3", "ext4"]
+- block:
+  - name: Install e2fsprogs for ext file system type
+    package:
+      name: e2fsprogs
+      state: present
+  - set_fact:
+      storage_packages: "{{ storage_packages + ['e2fsprogs'] }}"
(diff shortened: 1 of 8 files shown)


tasks/vg-default.yml
```diff
@@ -4,12 +4,19 @@
...
+    package:
+      name: lvm2
+      state: present
+  - set_fact:
+      storage_packages: "{{ storage_packages + ['lvm2'] }}"
+  when: "'lvm2' not in storage_packages and pool.type == 'lvm' and not ansible_check_mode"
+
+- name: Update ansible_facts after installing lvm2
+  setup:
+  run_once: true
```
dwlehman: Can you move the `setup` call into the `block` above?

timflannagan1: I tried that in the initial development stage. Unfortunately, I get the following error when I try to move `setup` into the task block above: 

```
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'pool.type == \"lvm\" and pool.name in ansible_facts.lvm.vgs' failed. The error was: error while evaluating conditional (pool.type == \"lvm\" and pool.name in ansible_facts.lvm.vgs): 'dict object' has no attribute 'lvm'\n\nThe error appears to have been in '/home/tim/Documents/storage/tasks/vg-default.yml': line 43, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n- block:\n    - name: collect list of current pvs\n      ^ here\n"}
```

tasks/fs-default.yml
```diff
@@ -6,19 +6,28 @@
   package:
     name: xfsprogs
     state: present
-  when: volume.fs_type == "xfs" and not ansible_check_mode
+  register: xfsprogs_installed
+  when: "volume.fs_type == 'xfs' and ['xfsprogs'] is not subset(ansible_facts.packages.keys()) \
+         and xfsprogs_installed is undefined and not ansible_check_mode"
```
dwlehman: Might want to add a check for `volume.state == 'present'` since we only need ext2 and xfs tools to create/resize -- not to remove.

timflannagan1: Updated. About a quarter of the time, the `setup` module was being skipped after installing `lvm2` when absent in the user's system. I moved the `lvm_installed` variable from a registered variable from the `package` command to a fact so the `setup` module will always run.

PR #9, opened 2018-11-07
Remove a partition when a storage_pool's state is set to absent and use_partitions is false/undefined.

### LINKED ISSUE #5: Playbook fails to remove any partitions from a storage_pool if use_partitions is undefined or false
* When the variable `use_partitions` is set to false, or it's undefined, then we encounter an error where we cannot remove a `storage_pool` that contains partitions, as the playbook tasks will interpret this as an `lvm` volume group and the partition(s) won't be removed.
* It's also worth noting that the playbook will continue to run and won't fail after failing to remove that pool.

timflannagan1:
* This only handles removing primary lvm partitions.
* Fixed #5

### DIFF
--- a/tasks/vg-default.yml
+++ b/tasks/vg-default.yml
@@ -86,9 +86,19 @@
     num_holders: "{{ ansible_devices[pool_disk_suffix].partitions[part_name].holders | length | bool }}"
   when: part_name is defined
 
-- name: Remove the partition on storage_pools disk if pool.state is absent
-  parted:
-    device: "/dev/{{ pool_disk_suffix }}"
-    state: "{{ pool.state }}"
-    number: 1
+- block:
+    - name: Remove the partition on storage_pools disk if pool.state is absent
+      parted:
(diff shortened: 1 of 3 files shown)


tasks/vg-default.yml
```diff
@@ -72,3 +72,33 @@
     - debug:
         msg: "Failed to wipe pv signatures."
   when: pool.type == 'lvm' and pool.state == "absent"
+
+#
+# Handle removing a partition when pool.state is absent and use_parititions is undefined/false
+#
+- name: Get the name of the pool's primary partition as applicable
```
dwlehman: It would be more robust if you were to iterate over `pool._orig_members` instead and remove any that are partitions as there could be multiple partitions across several disks.

timflannagan1: Ended up iterating over `pool.disks` instead as throwing the partition path (e.g. `/dev/sdb1') to `parted` won't change anything and will evaluate to `ok: [localhost]`.

PR #15, opened 2018-11-30
Need to use set_fact instead of register to track whether a package has been installed.

timflannagan1:
* This was found when running a playbook where the first role instantiation created an lvm group with xfs, and the second role just changed the fs_type to ext4. The playbook would fail as the system didn't have the mkfs.ext4 executable. This means that `e2fsprogs` task was skipped in the second run-through as the registered variable `e2fsprogs_installed` was interpreted as being defined in the previous role run-through despite being skipped initially.

### Playbook used:
```yaml
---
- hosts: localhost
  become: true
  vars:

  roles:
    - name: storage
      storage_pools:
        - name: foo
          disks: ['vdc']
          volumes:
            - name: test1
              size: 10g
              fs_type: 'xfs'
              mount_point: '/opt/test1'

    - name: storage
      storage_pools:
        - name: foo
          disks: ['vdc']
          volumes:
            - name: test1
              size: 10g
              fs_type: 'ext4'
              mount_point: '/opt/test1'
```

### DIFF
--- a/tasks/fs-default.yml
+++ b/tasks/fs-default.yml
@@ -2,30 +2,36 @@
 - name: Stat the final device file
   include_tasks: stat_device.yml
 
-- name: Install xfsprogs for xfs file system type
-  package:
-    name: xfsprogs
-    state: present
-  register: xfsprogs_installed
+- block:
+  - name: Install xfsprogs for xfs file system type
+    package:
+      name: xfsprogs
+      state: present
+  - set_fact:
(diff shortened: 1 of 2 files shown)


timflannagan1:
@dwlehman

timflannagan1:
Added `parted_installed` in the last force-push and the lv-default.yml, and vg-default.yml were already using the `set_fact` structure.

PR #17, opened 2018-12-05
Add various integration tests

timflannagan1:
* This would add multiple playbooks designing to test the interactions between role components.
* In each playbook, after each role is run, it's verified by including the task file `tests/verify-role-results.yml`, which is a driver that uses assert statements to compare the role results and the user's system results. These tests rely on the internal structures (`_storage_pools_list`, `_storage_volumes_list`) that are built up during role execution.


timflannagan1:
@dwlehman I believe I was waiting for you to review it, but let me pull the changes made to the upstream master in the last couple of months first then test it out.

pcahyna:
[citest pending]

pcahyna:
[citest bad]

pcahyna:
[citest bad]

tests/tests_change_mount.yml
```diff
@@ -0,0 +1,41 @@
...
+                mount_point: "{{ mount_location_before }}"
+
+    - include_tasks: verify-role-results.yml
+
+    - name: Change the file system signature on /dev/mapper/foo-test1
+      include_role:
+        name: storage
+      vars:
+        storage_pools:
+          - storage_pools:
```
timflannagan1: this needs to be removed

pcahyna: what part exactly?

timflannagan1: just the redundant `- storage_pools:` under the first `storage_pools:`

timflannagan1:
@pcahyna it appears all the tests are running successfully as long as there's enough disk storage for each environment

pcahyna:
@timflannagan1 indeed, which is great, thanks. We tested them locally as well.

yontalcar:
This looks weird:
```
TASK [Find unused disks in the system] *****************************************
task path: /tmp/tmp3jq95ok8/tests/get_unused_disk.yml:2
ok: [/cache/CentOS-6-x86_64-GenericCloud-1804_02.qcow2c] => {"changed": false, "disks": ["ram14", "ram15", "ram12", "ram13", "ram10", "ram11", "ram8", "ram9", "ram4", "ram5"]}
```
`unused_disks` are resolved as `/dev/ram*` instead of `/dev/vd*`

Also some tests failed because wipefs in rhel/centos 6 doesn't recognise  `--force` option (see https://github.com/karelzak/util-linux/commit/2968c3fc7388f88b8debe64d61d9785601c16436)

timflannagan1:
@yontalcar I'm seeing in the centos 6 guest.log that vda and vdb have `unknown partition table` for some reason, which is why it would be considered a used disk.

dwlehman:
Having an unknown partition table (aka "no partition table") is one of the criteria for being an unused disk. There must be something else going on.

yontalcar:
By quick glance onto `library/find_unused_disk.py` it returns (by default) first 10 found disk, so it doesn't even get to `/dev/vdX`. Also it looks that playbook uses first disk found, but we need disk with some reasonable size.

Other problem that I hit was that there is no systemd on rhel/centos 6

timflannagan1:
We could add a parameter to `library/find_unused_disk` that specifies the minimum disk size required and let  `module_utils.size.py` handle both the module parameter and the current disk size which Ansible facts reports.

library/find_unused_disk.py
```diff
@@ -27,7 +32,9 @@
   tasks:
     - name: run module
       find_unused_disk:
+        disk_size: '10g'
```
dwlehman: It would be nice if it were called `min_size` since that how it gets interpreted.

library/find_unused_disk.py
```diff
@@ -106,19 +112,16 @@ def run_module():
...
             if len(result['disks']) >= module.params['max_return']:
                 break
-        else:
-            disk_info = { 'name': disk,
-                          'partitions': no_partitions,
-                          'signature': no_signature(run_command, '/dev/' + disk),
-                          'holders': no_holders(disk),
-                          'can_open': can_open('/dev/' + disk)
-            }
-            result['used_disks'].append(disk_info)
```
dwlehman: When was this added? Is it just some debug artifacts?

timflannagan1: I was using it purely for debugging purposes for the centos/rhel 6 VMs and was erased in the most current commit. 

I'll have to rebase all these commits soon.

yontalcar:
[citest bad]
[citest pending]

timflannagan1:
[citest bad]

timflannagan1:
Squashed all the commits into the original commit, and updated the commit message to better reflect the changes made in this PR.

pcahyna:
> Unmount an existing filesystem of a device if we want to change the mount location, but keep the same fs-type.

mount(8) says that there is a  `mount --move olddir newdir` command to atomically move mounts.

Also, how is this fix related to "Adding various integration tests"? Could you please merge the integration tests and do further development in new branches?

timflannagan1:
The last two commits can be moved to a separate PR then. Once that's done, the PR should be in a good spot.

timflannagan1:
[citest pending]

timflannagan1:
@pcahyna @dwlehman It appears that adding the tags we talked about in the `tests/tests_change_disk*` playbooks ended up working and the tests are passing now.

pcahyna:
@timflannagan1 good, thanks!
I think you don't need to squash the last commit which adds the tags. This way it will be easy to revert when the tests are fixed.

timflannagan1:
@pcahyna alright I'll kept that commit, and squashed the other two into the original commit.

pcahyna:
@timflannagan1 @dwlehman is this PR good to merge now? Do not worry about test failures in the staging CI environment.

timflannagan1:
@pcahyna there's a small debug artifact in `tests/tests_change_disk_fs.yml`, but I can remove that in subsequent PRs. Otherwise, I think it's in a good spot.

timflannagan1:
Removed the debug task in `tests/tests_change_disk_fs.yml`, so this PR should be okay to merge now.

PR #19, opened 2018-12-11
Documentation update of modules

japokorn:
- modules should be now pep8 compliant (flake8)
- documentation formatting updated
- external commands are now properly checked for presence on the target
system


library/bsize.py
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/python
+# Copyright (c) 2017 Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
pcahyna: - should the modules be (c) Ansible Project? And why 2017?
- should they be GPL, when the rest of the role is MIT?

@timflannagan1 @dwlehman did you write those modules? What do you think?

Issue #22, opened 2019-02-05
Modifying the mount location on an existing disk device causes the playbook to fail


timflannagan1:
### Example of playbook that would reproduce this:
```
- name: Create a disk device mounted at '/opt/test1'
      include_role:
        name: storage
      vars:
        storage_volumes:
          - name: test1
            type: disk
            mount_point: '/opt/test1'
            disks: ['vdb']
    
    - name: Change the disk device mount location to '/opt/test2'
      include_role:
        name: storage
      vars:
        storage_volumes:
          - name: test1
            type: disk
            mount_point: '/opt/test2'
            disks: ['vdb']
```
### Output:
```
TASK [storage : Remove file system as needed] *********************************************************************************************************************************************************************
task path: /home/tim/Documents/storage/tasks/fs-default.yml:59
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: tim
<127.0.0.1> EXEC /bin/sh -c 'echo ~root && sleep 0'
<127.0.0.1> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /root/.ansible/tmp/ansible-tmp-1549388550.8278959-101630898191692 `" && echo ansible-tmp-1549388550.8278959-101630898191692="` echo /root/.ansible/tmp/ansible-tmp-1549388550.8278959-101630898191692 `" ) && sleep 0'
Using module file /usr/lib/python3.7/site-packages/ansible/modules/commands/command.py
<127.0.0.1> PUT /root/.ansible/tmp/ansible-local-9757nqfhq7kn/tmp1jn3g0hm TO /root/.ansible/tmp/ansible-tmp-1549388550.8278959-101630898191692/AnsiballZ_command.py
<127.0.0.1> EXEC /bin/sh -c 'chmod u+x /root/.ansible/tmp/ansible-tmp-1549388550.8278959-101630898191692/ /root/.ansible/tmp/ansible-tmp-1549388550.8278959-101630898191692/AnsiballZ_command.py && sleep 0'
<127.0.0.1> EXEC /bin/sh -c 'sudo -H -S -n -u root /bin/sh -c '"'"'echo BECOME-SUCCESS-ylcynifwzjqwmtbtcmikvotmzrvbrgag; /usr/bin/python3 /root/.ansible/tmp/ansible-tmp-1549388550.8278959-101630898191692/AnsiballZ_command.py'"'"' && sleep 0'
[...]

PR #23, opened 2019-03-22
WIP: Add workaround in tasks/fs-default.yml to handle same fs-type situations.

timflannagan1:
* The `tests/tests_change_disk*` playbooks are currently failing due to #22. 
* This change would remove the mount and /etc/fstab entry of a present disk device, with the intention of either re-formatting or modifying the mount location of that same disk.


timflannagan1:
Closing due to #24

PR #24, opened 2019-03-25
Add workaround to tasks/fs-default.yml to handle same fs-type situati…

timflannagan1:
…ons with storage-volumes.

### DIFF
--- a/tasks/fs-default.yml
+++ b/tasks/fs-default.yml
@@ -42,12 +42,25 @@
          and ['util-linux'] is not subset(ansible_facts.packages.keys()) and
          util_linux_installed is undefined and not ansible_check_mode"
 
-- name: unmount fs if we're going to reformat
+- name: Check if we are trying to format the same fs-type as the current device fs-type
+  set_fact:
+    same_fs_flag: True
+  when: volume.fs_type == volume._orig_fs_type
+
+- name: Unmount fs if we're going to reformat
   mount:
     src: "{{ volume._device }}"
     path: "{{ volume._orig_mount_point }}"
     state: unmounted
   when: volume._wipe and volume._orig_mount_point
+  changed_when: false
+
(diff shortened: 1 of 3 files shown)


tasks/fs-default.yml
```diff
@@ -42,14 +42,27 @@
          and ['util-linux'] is not subset(ansible_facts.packages.keys()) and
          util_linux_installed is undefined and not ansible_check_mode"
 
-- name: unmount fs if we're going to reformat
+- name: Check if we are trying to format the same fs-type as the current device fs-type
+  set_fact:
+    same_fs_flag: True
+  when: volume.fs_type == volume._orig_fs_type
```
timflannagan1: In `defaults/main.yml`, we specify `volume._orig_fs_type: ""`, and then update that scoped variable as needed in `tasks/volume-default.yml` when we run a blkid command to find a device's current fs-type, and a successful rc is returned. We set a flag to indicate we are passing the same fs-type as the one on an existing disk device we are trying to configure.

tasks/fs-default.yml
```diff
@@ -42,14 +42,27 @@
...
     state: unmounted
   when: volume._wipe and volume._orig_mount_point
   changed_when: false
 
+- name: Remove /etc/fstab entry if we are passing the same fs-type
+  mount:
+    src: "{{ volume._device }}"
+    path: "{{ volume._orig_mount_point }}"
+    state: absent
+  when: same_fs_flag is defined
```
timflannagan1: We gather a present device's UUID in `tasks/mount-default.yml`, which is used, if present, in `/etc/fstab` entries as the device identifier (instead of device path), but we first run any filesystem tasks before that. When we create the filesystem, it will assign a new UUID to the same device, and therefore produces two entries in fstab.

PR #25, opened 2019-04-09
Blivet backend WIP

dwlehman:
This is a rough WIP version. I want to get the CI running on it while I clean some things up.


pcahyna:
don't worry too much about the -staging failures and errors, that infrastructure is flaky.

pcahyna:
@dwlehman do you have any updates on this or is it ready for review?

dwlehman:
[citest all]

dwlehman:
[citest all]

dwlehman:
[citest bad]

dwlehman:
I would like to improve the structure of the code in `library/blivet.py` (to add some encapsulation around various volume types, but I think as of now the functionality is mostly in place for v1. One exception is partitioning which, although straightforward, will probably take a few hours to get right. It's worth noting that the tests involving partitions appear to not effectively test that partitions are actually used, which we'll want to fix.

Having said that, I think we need to focus on the functionality and the interfaces, and I can improve the internals later.

dwlehman:
As of now the support is there for a single partition that spans the whole disk, either as a volume or as a backing device for a pool. The former case is verified during testing, while the latter is not.

PR #26, opened 2019-05-09
syntax error with ansible 2.6

rhmk:
When running this role with ansible 2.6 an error is thrown "unknow operand". changing eq to equalto fixes the problem

### DIFF
--- a/tasks/volume-default.yml
+++ b/tasks/volume-default.yml
@@ -73,7 +73,7 @@
   when: volume._preexist
 
 - set_fact:
-    volume: "{{ volume|combine({'_orig_mount_point': ansible_facts.mounts|selectattr('device', 'eq', volume._device)|map(attribute='mount')|list|first|default('')}, recursive=True) }}"
+    volume: "{{ volume|combine({'_orig_mount_point': ansible_facts.mounts|selectattr('device', 'equalto', volume._device)|map(attribute='mount')|list|first|default('')}, recursive=True) }}"
 
 - set_fact:
     volume: "{{ volume|combine({'_wipe': volume._preexist and (not volume.fs_type or (volume._orig_fs_type and volume.fs_type != volume._orig_fs_type))}) }}"


pcahyna:
Thanks. This should not depend on the version of Ansible; it is a Jinja2 feature. "eq" as a synonym for "equalsto" was introduced in Jinja2 2.10: http://jinja.pocoo.org/docs/2.10/changelog/ (PR https://github.com/pallets/jinja/pull/665 ). I agree that we should use the name that is supported in older Jinja2 versions. How to test this in the CI? @tyll do you think it would be OK to use an older version of Jinja2 in the CI (would probably mean to downgrade the CI container)?

tyll:
> Thanks. This should not depend on the version of Ansible; it is a Jinja2 feature. "eq" as a synonym for "equalsto" was introduced in Jinja2 2.10: http://jinja.pocoo.org/docs/2.10/changelog/ (PR [pallets/jinja#665](https://github.com/pallets/jinja/pull/665) ). I agree that we should use the name that is supported in older Jinja2 versions. How to test this in the CI? @tyll do you think it would be OK to use an older version of Jinja2 in the CI (would probably mean to downgrade the CI container)?

If we want to control the versions, we should probably install them using pip instead of using RPMs. With RPMs we have only limited flexibility with downgrading packages and it is very likely that once we update the container image to Fedora 30, we will only have Jina 2.10 or newer as RPM.

pcahyna:
Travis on the other hand has Jinja2 v. 2.8, it seems.

pcahyna:
Hello, thanks for the patch.
I've just merged #34, which is a complete rewrite (uses a module written in Python and based on blivet), and your changes no longer apply. I've preserved the old version as the **prototype-0-1** branch and applied your PR to it, instead to master.

pcahyna:
Even `eq` is not supported in Jinja 2.7.2 that is in RHEL7 . We need to use `==`. @yontalcar

pcahyna:
Ideally the prototype-0-1 branch should be updated because it suffers from the same problem as in #49

PR #31, opened 2019-07-16
WIP: Test validation rewrite

dwlehman:


tests/test-verify-volume-mount.yml
```diff
@@ -0,0 +1,36 @@
+---
+- name: Set some facts
+  set_fact:
+    storage_test_mount_device_matches: "{{ ansible_mounts|selectattr('device', 'equalto', current_volume._device)|list }}"
+    storage_test_mount_point_matches: "{{ ansible_mounts|selectattr('mount', 'equalto', current_volume.mount_point|list }}"
+    storage_test_mount_expected_match_count: "{{ 1 if {{ current_volume.state == 'present' and current_volume.mount_point else 0|int }}"
```
pcahyna: Shouldn't you reload facts using the setup module before doing this? Otherwise `ansible_mounts` may contain stale data (collected at the beginning of the playbook). See https://github.com/linux-system-roles/selinux/blob/6cd1ec8fdebdb92a789b14e5a44fe77f0a3d8ecd/tests/selinux_test_transitions.yml#L11. Or is that the intent?

tests/verify-role-results.yml
```diff
@@ -11,8 +8,36 @@
...
+
+#
+# Clean up.
+#
+- name: Clean up variable namespace
+  set_fact:
+    storage_test_fstab: null
+    storage_test_blkinfo: null
+    current_pool: null
+    current_volume: null
```
pcahyna: good idea, since there is no way to unset a variable or fact and using the task file multiple times from the same playbook would otherwise yield unexpected results. Would it work to register `storage_test_blkinfo` always, instead of only if `storage_test_blkinfo is not defined or not storage_test_blkinfo`? This could be an alternative to setting it to `null`.

tests/test-verify-volume-mount.yml
dwlehman: The last thing we do in `tasks/main-blivet.yml` is run `setup` to pick up changes we made.

pcahyna: That's an interesting approach, maybe we should do that in the other roles too (network and selinux in particular)

tests/verify-role-results.yml
dwlehman: The intent is to collect the information once for all devices per test/validation run. Since I'm collecting the info in a tasks file that applies to a single volume (`test-verify-volume.yml`) I want to do so conditionally to prevent re-running it unnecessarily. I could collect the info at the top level, but it would be less clear in `verify-role-results.yml` what the information is being collected for. So I think this could be addressed either with a comment in `test-verify-volume.yml` or by moving the invocation of `blockdev_info` up into `verify-role-results.yml`.

I also plan to abstract the test-verification-specific facts (probably into a single dict) so that the top-level can reset them all without having to know what facts have been defined/used.

dwlehman:
[citest]

tasks/main-blivet.yml
```diff
@@ -82,6 +82,22 @@
...
+    disklabel_type: "{{ storage_disklabel_type }}"
+  register: package_info
+
+- name: make sure required packages are installed
+  package:
+    name: "{{ pkg_name }}"
+    state: present
+  loop: "{{ package_info.packages }}"
+  loop_control:
+    loop_var: pkg_name
```
pcahyna: the package module can accept a list, so you don't need to iterate. See the 3rd item in https://github.com/oasis-roles/meta_standards#ansible-best-practices

dwlehman: Ah, good. I did not like having to do them one at a time, but the module docs for `package` don't mention passing a list.

pcahyna: True, it is undocumented. But I don't think it will go away. We have more places using it already.

dwlehman: Fixed in latest commit (f94da5a).

tasks/main-blivet.yml
```diff
@@ -82,6 +82,22 @@
 - debug:
     var: _storage_volumes
 
+- name: get required packages
+  blivet:
```
pcahyna: You forgot to specify `packages_only` here, it seems.

library/blivet.py
```diff
@@ -700,6 +702,11 @@ def action_dict(action):
             module.fail_json(msg=str(e), **result)
 
     scheduled = b.devicetree.actions.find()
+    result['packages'] = b.packages[:]
+
+    if module.params['packages_only']:
+        module.exit_json(**result)
+
```
pcahyna: This won't work because if the packages are unavailable, we have already failed and exited after the manage_pool call (which tracebacks) a few lines above.

PR #32, opened 2019-07-22
Fix removal of filesystem on every run for non-LVM volumes

gadgetmg:
Currently, `volume._remove` always comes back as `true` when working with non-LVM volumes. This is because `pool` being undefined will always make the conditional true.

I assume that the intended behavior is for a volume to be removed when either is is configured as absent or it's parent pool is absent.

I've only tested this with my use-case of managing partitions without LVM, but this *shouldn't* break anything there.

### DIFF
--- a/tasks/volume-default.yml
+++ b/tasks/volume-default.yml
@@ -79,7 +79,7 @@
     volume: "{{ volume|combine({'_wipe': volume._preexist and (not volume.fs_type or (volume._orig_fs_type and volume.fs_type != volume._orig_fs_type))}) }}"
 
 - set_fact:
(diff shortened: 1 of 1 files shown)


dwlehman:
[citest]

pcahyna:
[citest commit:57258a06908fcbdc65bb7632743419e6cc73d222]

pcahyna:
@dwlehman is the problem present in the blivet rewrite as well?

dwlehman:
I don't think this problem exists in the blivet rewrite. The (passing) idempotence tests should be confirmation.

pcahyna:
Hello, thanks for the patch.
I've just merged #34, which is a complete rewrite (uses a module written in Python and based on blivet), and your changes no longer apply. I've preserved the old version as the **prototype-0-1** branch and applied your PR to it, instead to master.

PR #33, opened 2019-08-06
fixes for the blivet branch 

pcahyna:
Includes the code in #28 (blivet refactor) and @vcrhonek 's Molecule setup ( https://github.com/vcrhonek/storage/tree/molecule )

PR #28, opened 2019-06-05
WIP: blivet module refactor

dwlehman:
This is the beginnings of a blivet module that will be easier to maintain over time than the one in #25.

PR #34, opened 2019-08-12
Fixes for the test validation rewrite branch

pcahyna:
Includes the code in #28 & #31 (blivet refactor) and @vcrhonek 's Molecule setup ( https://github.com/vcrhonek/storage/tree/molecule )
Supersedes #33


tasks/main.yml
```diff
@@ -1,4 +1,11 @@
 ---
+- name: Set version specific variables
+  include_vars: "{{ item }}"
+  with_first_found:
+    - "{{ ansible_distribution }}-{{ ansible_distribution_major_version }}.yml"
+    - "{{ ansible_distribution }}.yml"
+    - "{{ ansible_os_family }}.yml"
+
```
pcahyna: @t-woerner has a similar but better solution: https://github.com/freeipa/ansible-freeipa/blob/master/roles/ipaserver/tasks/main.yml

PR #35, opened 2019-08-13
Improve the README

pcahyna:


README.md
```diff
@@ -25,7 +25,8 @@ This specifies the type of pool to manage.
 Valid values for `type`: `lvm`.
 
 ##### `disks`
-This specifies the set of disks to use as backing storage for the pool.
+A list which specifies the set of disks to use as backing storage for the pool
+(a `/dev/` or `/dev/mapper/` prefix is prepended if the name is not an absolute path).
```
dwlehman: I think this is underselling what we can do. How about something like "(supported identifiers include: device node, device node basename, /dev/disk/ symlink)"

dwlehman: Alternatively, we can provide an example of each form of identifier:

- sda
- /dev/sda
- mpathb
- /dev/mapper/mpathb
- /dev/disk/by-id/wwn-0x5000c5005bc37f3f

README.md
```diff
@@ -79,18 +83,20 @@ Example Playbook
   roles:
     - name: storage
       storage_pools:
-        - name: "{{ app_name }}"
-          disks: "{{ app_data_wwns }}"
+        - name: app
+          disks:
+            - /dev/sdb
+            - /dev/sdc
```
dwlehman: I would personally emphasize the convenience we provide by omitting the '/dev/' prefix on these.

README.md
pcahyna: Is there a particular capability to handle /dev/disk/ symlinks? I would suppose that if the role can handle device nodes, it will also handle symlinks to device nodes without any additional code, the point of symlinks is to behave like their target files. Or is there more to it or anything special about /dev/disk symlinks? (/dev/mapper device nodes are usually also symlinks.)

dwlehman: It does not handle arbitrary symlinks, but it seemed worth pointing out /dev/disk symlinks because they are useful in that some of the use stable identifiers that are not subject to change across reboots, etc.

pcahyna: Out of curiosity, why doesn't it handle arbitrary symlinks?

dwlehman: There hasn't been any use case or request for resolving arbitrary symlinks to block devices.

Issue #42, opened 2019-09-20
Document default to destructive behavior


tabowling:
@dwlehman  reminder that we need to clearly communicate to end users via upstream documentation that the current behavior is to default to destructively overwriting existing and conflicting volume groups and filesystems.

We will need to file a separate issue to default to safely preserving data and having the role fail in the event of conflicts while providing the user some option to specify that they desire overwriting and destroying existing state.  But for now we simply need to clearly and emphatically communicate the behavior.

pcahyna:
I would like to know more details about the cases where the behaviour is destructive. @dwlehman can you please elaborate a bit?

PR #43, opened 2019-09-25
Add flag to prevent implicit removal of existing device/fs.

dwlehman:
This flag, `storage_safe_mode`, will prevent the implicit/automatic removal of existing file systems or device stacks. In other words, with this flag set, nothing will be removed except pools and volumes whose state is defined as `absent`.

This is related to #42

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -3,6 +3,7 @@
 storage_provider: "blivet"
 storage_use_partitions: null
 storage_disklabel_type: null  # leave unset to allow the role to select an appropriate label type
+storage_safe_mode: true  # fail instead of implicitly/automatically removing devices or formatting
 
 storage_pool_defaults:
   state: "present"
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -31,6 +31,10 @@ options:
     disklabel_type:
         description:
             - disklabel type string (eg: 'gpt') to use, overriding the built-in logic in blivet
+    safe_mode:
+        description:
+            - boolean indicating that we should fail rather than implicitly/automatically
+              removing devices or formatting
 
 author:
     - David Lehman (dlehman@redhat.com)
@@ -116,6 +120,7 @@ if BLIVET_PACKAGE:
 
 use_partitions = None  # create partitions on pool backing device disks?
 disklabel_type = None  # user-specified disklabel type
+safe_mode = None       # do not remove any existing devices or formatting
 
 
 class BlivetAnsibleError(Exception):
@@ -163,7 +168,6 @@ class BlivetVolume(object):
                          label=self._volume['fs_label'],
                          options=self._volume['fs_create_options'])
         if not fmt.supported or not fmt.formattable:
-            # FAIL: fs type tools are not available
             raise BlivetAnsibleError("required tools for file system '%s' are missing" % self._volume['fs_type'])
 
         return fmt
@@ -189,7 +193,6 @@ class BlivetVolume(object):
         try:
             size = Size(self._volume['size'])
         except Exception:
-            # FAIL: invalid size specification
             raise BlivetAnsibleError("invalid size specification for volume '%s': '%s'" % (self._volume['name'], self._volume['size']))
 
         if size and self._device.resizable and self._device.size != size:
@@ -197,13 +200,11 @@ class BlivetVolume(object):
                 self._device.format.update_size_info()
 
             if not self._device.min_size <= size <= self._device.max_size:
-                # FAIL: resize to specified size not possible
                 raise BlivetAnsibleError("volume '%s' cannot be resized to '%s'" % (self._volume['name'], size))
 
             try:
                 self._blivet.resize_device(self._device, size)
             except ValueError as e:
-                # FAIL: resize not possible
                 raise BlivetAnsibleError("volume '%s' cannot be resized from %s to %s: %s" % (self._device.name,
                                                                                               self._device.size,
                                                                                               size, str(e)))
@@ -213,6 +214,9 @@ class BlivetVolume(object):
         fmt = self._get_format()
         if self._device.format.type == fmt.type:
             return
+
+        if safe_mode and (self._device.format.type is not None or self._device.format.name != get_format(None).name):
+            raise BlivetAnsibleError("cannot remove existing formatting on volume '%s' in safe mode" % self._volume['name'])
 
         if self._device.format.status:
             self._device.format.teardown()
@@ -255,6 +259,17 @@ class BlivetDiskVolume(BlivetVolume):
     def _type_check(self):
         return self._device.is_disk
 
+    def _look_up_device(self):
+        super(BlivetDiskVolume, self)._look_up_device()
+        if not self._get_device_id():
+            raise BlivetAnsibleError("no disks specified for volume '%s'" % self._volume['name'])
+        elif not isinstance(self._volume['disks'], list):
+            raise BlivetAnsibleError("volume disks must be specified as a list")
+
+        if self._device is None:
+            raise BlivetAnsibleError("unable to resolve disk specified for volume '%s' (%s)" % (self._volume['name'], self._volume['disks']))
+
+
 
 class BlivetPartitionVolume(BlivetVolume):
     def _type_check(self):
@@ -273,21 +288,18 @@ class BlivetPartitionVolume(BlivetVolume):
             parent = self._blivet.devicetree.resolve_device(self._volume['pool'])
 
         if parent is None:
-            # FAIL: failed to find pool
             raise BlivetAnsibleError("failed to find pool '%s' for volume '%s'" % (self._blivet_pool['name'], self._volume['name']))
 
         size = Size("256 MiB")
         try:
             device = self._blivet.new_partition(parents=[parent], size=size, grow=True, fmt=self._get_format())
         except Exception:
-            # FAIL: failed to instantiate volume device
             raise BlivetAnsibleError("failed set up volume '%s'" % self._volume['name'])
 
         self._blivet.create_device(device)
         try:
             do_partitioning(self._blivet)
         except Exception:
-            # FAIL: partition allocation failed: not enough space?
             raise BlivetAnsibleError("partition allocation failed for volume '%s'" % self._volume['name'])
 
         self._device = device
@@ -303,18 +315,15 @@ class BlivetLVMVolume(BlivetVolume):
 
         parent = self._blivet_pool._device
         if parent is None:
-            # FAIL: failed to find pool
             raise BlivetAnsibleError("failed to find pool '%s' for volume '%s'" % (self._blivet_pool['name'], self._volume['name']))
 
         try:
             size = Size(self._volume['size'])
         except Exception:
-            # FAIL: invalid size specification
             raise BlivetAnsibleError("invalid size '%s' specified for volume '%s'" % (self._volume['size'], self._volume['name']))
 
         fmt = self._get_format()
         if size > parent.free_space:
-            # FAIL: volume size greater than pool free space
             raise BlivetAnsibleError("specified size for volume '%s' exceeds available space in pool '%s' (%s)" % (size,
                                                                                                                    parent.name,
                                                                                                                    parent.free_space))
@@ -323,7 +332,6 @@ class BlivetLVMVolume(BlivetVolume):
             device = self._blivet.new_lv(name=self._volume['name'],
                                          parents=[parent], size=size, fmt=fmt)
         except Exception:
-            # FAIL: failed to create volume
             raise BlivetAnsibleError("failed to set up volume '%s'" % self._volume['name'])
 
         self._blivet.create_device(device)
@@ -391,8 +399,7 @@ class BlivetPool(object):
     def _look_up_disks(self):
         """ Look up the pool's disks in blivet's device tree. """
         if not self._pool['disks']:
-            # FAIL: no disks specified for pool
-            raise BlivetAnsibleError("no disks specified for pool '%s'" % self._pool['name'])  # sure about this one?
+            raise BlivetAnsibleError("no disks specified for pool '%s'" % self._pool['name'])
         elif not isinstance(self._pool['disks'], list):
             raise BlivetAnsibleError("pool disks must be specified as a list")
 
@@ -403,7 +410,6 @@ class BlivetPool(object):
                 disks.append(device)
 
         if self._pool['disks'] and not disks:
-            # FAIL: failed to find any disks
             raise BlivetAnsibleError("unable to resolve any disks specified for pool '%s' (%s)" % (self._pool['name'], self._pool['disks']))
 
         self._disks = disks
@@ -428,8 +434,11 @@ class BlivetPool(object):
         """ Schedule actions as needed to ensure pool member devices exist. """
         members = list()
         for disk in self._disks:
-            if not disk.isleaf:
-                self._blivet.devicetree.recursive_remove(disk)
+            if not disk.isleaf or disk.format.type is not None:
+                if safe_mode:
+                    raise BlivetAnsibleError("cannot remove existing formatting and/or devices on disk '%s' (pool '%s') in safe mode" % (disk.name, self._pool['name']))
+                else:
+                    self._blivet.devicetree.recursive_remove(disk)
 
             if use_partitions:
                 label = get_format("disklabel", device=disk.path)
@@ -446,7 +455,6 @@ class BlivetPool(object):
             try:
                 do_partitioning(self._blivet)
             except Exception:
-                # FAIL: problem allocating partitions for pool backing devices
                 raise BlivetAnsibleError("failed to allocation partitions for pool '%s'" % self._pool['name'])
 
         return members
@@ -490,7 +498,11 @@ class BlivetPartitionPool(BlivetPool):
     def _create(self):
         if self._device.format.type != "disklabel" or \
            self._device.format.label_type != disklabel_type:
-            self._blivet.devicetree.recursive_remove(self._device, remove_device=False)
+            if safe_mode:
+                raise BlivetAnsibleError("cannot remove existing formatting and/or devices on disk '%s' "
+                                         "(pool '%s') in safe mode" % (self._device.name, self._pool['name']))
+            else:
+                self._blivet.devicetree.recursive_remove(self._device, remove_device=False)
 
             label = get_format("disklabel", device=self._device.path, label_type=disklabel_type)
             self._blivet.format_device(self._device, label)
@@ -503,7 +515,6 @@ class BlivetLVMPool(BlivetPool):
     def _get_format(self):
         fmt = get_format("lvmpv")
         if not fmt.supported or not fmt.formattable:
-            # FAIL: lvm tools are not available
             raise BlivetAnsibleError("required tools for managing LVM are missing")
 
         return fmt
@@ -516,7 +527,6 @@ class BlivetLVMPool(BlivetPool):
         try:
             pool_device = self._blivet.new_vg(name=self._pool['name'], parents=members)
         except Exception:
-            # FAIL: failed to instantiate pool device
             raise BlivetAnsibleError("failed to set up pool '%s'" % self._pool['name'])
 
         self._blivet.create_device(pool_device)
@@ -660,6 +670,7 @@ def run_module():
         volumes=dict(type='list'),
         packages_only=dict(type='bool', required=False, default=False),
         disklabel_type=dict(type='str', required=False, default=None),
+        safe_mode=dict(type='bool', required=False, default=True),
         use_partitions=dict(type='bool', required=False, default=True))
 
     # seed the result dict in the object
@@ -691,6 +702,9 @@ def run_module():
 
     global use_partitions
     use_partitions = module.params['use_partitions']
+
+    global safe_mode
+    safe_mode = module.params['safe_mode']
 
     b = Blivet()
     b.reset()
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -218,7 +218,8 @@ class BlivetVolume(object):
         if self._device.format.type == fmt.type:
             return
 
-        if safe_mode and (self._device.format.type is not None or self._device.format.name != get_format(None).name):
+        if safe_mode and (self._device.format.type is not None or self._device.format.name != get_format(None).name) and \
+           not packages_only:
             raise BlivetAnsibleError("cannot remove existing formatting on volume '%s' in safe mode" % self._volume['name'])
 
         if self._device.format.status and not packages_only:
@@ -438,7 +439,7 @@ class BlivetPool(object):
         members = list()
         for disk in self._disks:
             if not disk.isleaf or disk.format.type is not None:
-                if safe_mode:
+                if safe_mode and not packages_only:
                     raise BlivetAnsibleError("cannot remove existing formatting and/or devices on disk '%s' (pool '%s') in safe mode" % (disk.name, self._pool['name']))
                 else:
                     self._blivet.devicetree.recursive_remove(disk)
@@ -501,7 +502,7 @@ class BlivetPartitionPool(BlivetPool):
     def _create(self):
         if self._device.format.type != "disklabel" or \
            self._device.format.label_type != disklabel_type:
-            if safe_mode:
+            if safe_mode and not packages_only:
                 raise BlivetAnsibleError("cannot remove existing formatting and/or devices on disk '%s' "
                                          "(pool '%s') in safe mode" % (self._device.name, self._pool['name']))
             else:
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -534,7 +534,7 @@ class BlivetLVMPool(BlivetPool):
 
 
 _BLIVET_POOL_TYPES = {
-    "disk": BlivetPartitionPool,
+    "partition": BlivetPartitionPool,
     "lvm": BlivetLVMPool
 }
 
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -121,6 +121,7 @@ if BLIVET_PACKAGE:
 use_partitions = None  # create partitions on pool backing device disks?
 disklabel_type = None  # user-specified disklabel type
 safe_mode = None       # do not remove any existing devices or formatting
+packages_only = None   # only set things up enough to get a list of required packages
 
 
 class BlivetAnsibleError(Exception):
@@ -211,6 +212,8 @@ class BlivetVolume(object):
 
     def _reformat(self):
         """ Schedule actions as needed to ensure the volume is formatted as specified. """
+        global packages_only
+
         fmt = self._get_format()
         if self._device.format.type == fmt.type:
             return
@@ -218,7 +221,7 @@ class BlivetVolume(object):
         if safe_mode and (self._device.format.type is not None or self._device.format.name != get_format(None).name):
             raise BlivetAnsibleError("cannot remove existing formatting on volume '%s' in safe mode" % self._volume['name'])
 
-        if self._device.format.status:
+        if self._device.format.status and not packages_only:
             self._device.format.teardown()
         self._blivet.format_device(self._device, fmt)
 
@@ -705,6 +708,9 @@ def run_module():
 
     global safe_mode
     safe_mode = module.params['safe_mode']
+
+    global packages_only
+    packages_only = module.params['packages_only']
 
     b = Blivet()
     b.reset()
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -38,7 +38,7 @@
     _storage_vols_no_defaults: "{{ _storage_vols_no_defaults|default([]) }} + [{{ item.1 }}]"
     _storage_vol_defaults: "{{ _storage_vol_defaults|default([]) }} + [{{ storage_volume_defaults }}]"
     _storage_vol_pools: "{{ _storage_vol_pools|default([]) }} + ['{{ item.0.name }}']"
-  loop: "{{ _storage_pools|subelements('volumes') }}"
+  loop: "{{ _storage_pools|subelements('volumes', skip_missing=true) }}"
   when: storage_pools is defined
 
 - name: Apply defaults to pools and volumes [3/6]
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -105,6 +105,7 @@
     volumes: "{{ _storage_volumes }}"
     use_partitions: "{{ storage_use_partitions }}"
     disklabel_type: "{{ storage_disklabel_type }}"
+    safe_mode: "{{ storage_safe_mode }}"
   register: blivet_output
 
 - debug:
--- a/tests/tests_disk_errors.yml
+++ b/tests/tests_disk_errors.yml
@@ -73,7 +73,6 @@
           include_role:
             name: storage
           vars:
-            storage_safe_mode: true
             storage_volumes:
               - name: test1
                 type: disk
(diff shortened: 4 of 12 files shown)


tests/tests_lvm_errors.yml
```diff
@@ -149,3 +149,89 @@
...
+          vars:
+            storage_safe_mode: true
+            storage_pools:
+              - name: testpool1
+                type: lvm
+                disks: "{{ unused_disks }}"
+                volumes:
+                  - name: testvol1
+                    fs_type: 'ext4'
+                    size: '2g'
```
pcahyna: Why should safe mode prevent resizing of a volume? Is it a destructive operation?

defaults/main.yml
```diff
@@ -3,6 +3,7 @@
 storage_provider: "blivet"
 storage_use_partitions: null
 storage_disklabel_type: null  # leave unset to allow the role to select an appropriate label type
+storage_safe_mode: false  # do not remove devices as needed to make space for new volumes
```
pcahyna: Does the role remove existing volumes to make space for new ones? I.e. if you have a pool (VG) which is completely used by a volume oldlv, and you try to create a volume newlv, does the role remove oldlv to create space for newlv?

dwlehman: No.

tests/tests_lvm_errors.yml
dwlehman: If it goes wrong it can be ;)

defaults/main.yml
pcahyna: That's good to hear, but the comment above sounds like it does.

dwlehman: ...but I see your point. I'll change the wording. I wrote that comment before I wrote the actual code changes.

tests/tests_lvm_errors.yml
pcahyna: Isn't it the case for every storage operation? Is the risk particularly high for resizing?
For me resizing is a common operation and if it forces me to turn safe mode off, it just increases the likelihood that I will overwrite some valid data by accident. (Yes, if resizing is especially risky for some reason, it should be made conditional on safe mode being off, but this sounds like a deeper bug in the storage stack in the first place.)

defaults/main.yml
```diff
@@ -3,6 +3,7 @@
 storage_provider: "blivet"
 storage_use_partitions: null
 storage_disklabel_type: null  # leave unset to allow the role to select an appropriate label type
+storage_safe_mode: false  # fail instead of implicitly/automatically removing devices or formatting
```
pcahyna: Shouldn't the default be true? @tabowling what do you think?

tests/tests_lvm_errors.yml
dwlehman: The only difference is the possibility of losing pre-existing data you wanted to keep. I am open to dropping resize from this patch if the consensus is that it's not appropriate. I was just trying to make it really safe for the people who are into safety. (I'm not generally one of them.)

pcahyna: Sorry, I still don't get it. Why should resizing result in losing pre-existing data one wants to keep? When I use resize2fs or xfs_growfs, I don't consider them as unsafe operations.

dwlehman: The difference is this: if you are removing a device, you don't care about the data, so if the operation fails it does not matter. If you are creating a device there is no data to worry about losing if the operation goes awry. However, if you are resizing and things go sideways there is data involved that you presumably wanted to keep -- otherwise you would probably not be resizing.

pcahyna: I would say those are two very different meanings of "safe". When the role removes existing filesystems and physical volumes, it is sensitive to user errors. storage_safe_mode then protects users from their mistakes - when they accidentally asked for a filesystem to be created on a disk which contains valuable data instead of on an empty one.
In case of resizing, the only thing this setting protects from is a bug in the storage stack. Then, if we do not trust the storage stack enough, the functionality to resize filesystems should simply not be there, and if we trust it, it should not be governed by storage_safe_mode.
Given this difference, I really prefer storage_safe_mode to guard against overwrites of existing data only.

pcahyna: I.e. require `storage_safe_mode: false` mostly in situations where various management utilities already require some kind of --force flag.

dwlehman: That sounds fine to me: `storage_safe_mode` is about preventing users from accidentally removing their existing data. I will update the code and the descriptions accordingly.

pcahyna: @dwlehman thanks, and what is your opinion on enabling it by default? Otherwise there is a risk that the user will learn about in only when it is too late.

pcahyna:
@dwlehman it still silently destroys data in some situations. Consider a filesystem created with this playbook
```yaml
- hosts: all
  roles:
    - name: linux-system-roles.storage
      storage_volumes:
        - name: barefs
          type: disk
          disks:
            - /dev/vdc
          fs_type: xfs
          mount_point: /mnt/barefs
```
and then changed with this playbook
```yaml
- hosts: all
  roles:
    - name: linux-system-roles.storage
      storage_volumes:
        - name: barefs
          type: disk
          disks:
            - /dev/vdc
          fs_type: ext2
          mount_point: /mnt/barefs
```
(note the fs_type change), the filesystem is removed and recreated.

dwlehman:
On Thu, 2019-10-03 at 10:21 -0700, pcahyna wrote:
> @dwlehman it still silently destroys data in some situations.
> Consider a filesystem created with this playbook
> - hosts: all
>   roles:
>     - name: linux-system-roles.storage
>       storage_volumes:
>         - name: barefs
>           type: disk
>           disks:
>             - /dev/vdc
>           fs_type: xfs
>           mount_point: /mnt/barefs
> and then changed with this playbook
> - hosts: all
>   roles:
>     - name: linux-system-roles.storage
>       storage_volumes:
>         - name: barefs
>           type: disk
>           disks:
>             - /dev/vdc
>           fs_type: ext2
>           mount_point: /mnt/barefs
> (note the fs_type change), the filesystem is removed and recreated.

Yes, something must be wrong. Making safe mode the default should have
broken several of the existing integration tests.

pcahyna:
By the way, changing type from ext2 to ext3 is performed destructively as well, is it necessary? I would think that adding the journal inode and setting the HAS_JOURNAL flag should be easy.

pcahyna:
@dwlehman Are the known problems fixed now with the latest commits and the PR ready from your side, please?

dwlehman:
Yes, I think it's all set. FWIW there is a test case now for the last problem you found (in tests/tests_disk_errors.yml).

dwlehman:
Switching from ext2->ext3 is no different from ext2->xfs as of now. I'm a big fan of consistency. We can talk about putting in a special case for ext2->ext3, but it's not related to this pull request IMO.

pcahyna:
Concerning type switching, I found another surprise: if I omit `type:`, the volume gets reformatted with the default type (xfs). Wouldn't it be more reasonable to keep the existing filesystem?

dwlehman:
I'm not sure about that. I can see a case for either behavior. I have a somewhat difficult time imagining a user omitting `fs_type` and having a specific expectation about the result. Probably some people will complain whichever way we choose to interpret it.

pcahyna:
The scenario I am imagining is that someone will take a storage configuration created before by some other tools (anaconda, manually...), already used for storing data, and they will start using the storage role to manage it - change its properties, most likely size. It can be unexpected to see the the storage role reformatting your volumes (or at least trying to remove them, with the changes in this PR) and recreating them just because you did specify only the size (that you want to change) and not the filesystem type (or any other attribute that should be left unchanged).

library/blivet.py
```diff
@@ -660,6 +673,7 @@ def run_module():
         volumes=dict(type='list'),
         packages_only=dict(type='bool', required=False, default=False),
         disklabel_type=dict(type='str', required=False, default=None),
+        safe_mode=dict(type='bool', required=False, default=False),
```
pcahyna: Would it be easy to set safe_mode individually per pool or volume?
My reasoning is that one may need to reformat a disk. One then needs to set storage_safe_mode=false globally, and if one is managing the complete storage configuration in one call to the role, a mistake in an unrelated part of the configuration variable can then have unintended destructive consequences. It would be then safer to specify safe_mode only on the disks one needs to reformat.
(The tricky part is how to specify safe_mode for physical disks, which are provided in the "disks" arrays.)

dwlehman: How is reformatting a disk different from removing a file system in terms of being "unsafe" or "destructive"? Perhaps you are talking about _formatting_ an unformatted disk, which would not be prevented by `storage_safe_mode`.

pcahyna: I am talking about formatting a disk which previously had some data on it (to be discarded on purpose). This should require storage_safe_mode=false. Setting it does then remove the protection against mistakes in other parts of the configuration.

pcahyna:
> Yes, I think it's all set. FWIW there is a test case now for the last problem you found (in tests/tests_disk_errors.yml).

Sorry, but it still removes data for me. Consider this playbook (very similar to the two that I pasted above)
```yaml
- hosts: all
  roles:
    - name: linux-system-roles.storage
      vars:
        storage_safe_mode: true
        storage_volumes:
          - name: barefs
            type: disk
            disks:
              - /dev/vdc
            fs_type: xfs
            mount_point: /mnt/barefs
  tasks:
    - name: create a file
      file:
        path: /mnt/barefs/foo
        state: touch
- hosts: all
  roles:
    - name: linux-system-roles.storage
      vars:
        storage_safe_mode: true
        storage_volumes:
          - name: barefs
            type: disk
            disks:
              - /dev/vdc
            fs_type: ext2
            mount_point: /mnt/barefs
  tasks:
    - name: stat the file
      stat:
        path: /mnt/barefs/foo
      register: stat_r
    - name: assert file presence
      assert:
        that:
          stat_r.stat.isreg is defined and stat_r.stat.isreg
```
It should fail in the second role application, but in reality it fails only in the last assert that verifies that data have not been destroyed. I don't know how it is possible that your newly added test passes, but I see that you don't actually test for data not having been destroyed, only for role output.

tests/tests_disk_errors.yml
```diff
@@ -22,3 +30,77 @@
...
+                type: disk
+                fs_type: 'ext3'
+                disks: "{{ unused_disks }}"
+
+        - name: Verify the output
+          assert:
+            that: "{{ blivet_output.failed and
+                      blivet_output.msg|regex_search('cannot remove existing formatting on volume.*in safe mode') and
+                      not blivet_output.changed }}"
+            msg: "Unexpected behavior w/ existing data on specified disks"
```
pcahyna: I would say that this test is grossly insufficient considering the seriousness of the condition that it checks. What if the role destroys data and then sets blivet_output according to expectations?

tests/tests_disk_errors.yml
```diff
@@ -22,3 +30,77 @@
...
+          include_role:
+            name: storage
+          vars:
+            storage_volumes:
+              - name: test1
+                type: disk
+                disks: "{{ unused_disks }}"
+                present: false
+
+      ignore_errors: yes
```
pcahyna: @dwlehman How are asserts supposed to do anything useful in a block that has `ignore_errors: yes` ?

tests/tests_disk_errors.yml
```diff
@@ -22,3 +30,77 @@
...
+
+        - name: Create a partition pool on the disk already containing a file system w/o safe_mode
+          include_role:
+            name: storage
+          vars:
+            storage_safe_mode: false
+            storage_pools:
+              - name: foo
+                disks: "{{ unused_disks }}"
+                type: partition
```
pcahyna: This does not work.
```
TASK [storage : Apply defaults to pools and volumes [2/6]] *********************
task path: /home/pcahyna/linux-system-roles/storage/tasks/main-blivet.yml:36
fatal: [/home/pcahyna/linux-system-roles-testing/local/rhel-guest-image-7.7-200.x86_64.qcow2]: FAILED! => {"msg": "could not find 'volumes' key in iterated item {'state': 'present', 'type': 'partition', 'name': 'foo', 'disks': ['vdc']}"}
```
It is arguably a bug that one can not create an empty pool (partition or other). But until this is fixed, tests should take it into account.

tests/tests_disk_errors.yml
```diff
@@ -22,3 +30,77 @@
...
+
+        - name: Try to create a partition pool on the disk already containing a file system in safe_mode
+          include_role:
+            name: storage
+          vars:
+            storage_safe_mode: true
+            storage_pools:
+              - name: foo
+                disks: "{{ unused_disks }}"
+                type: partition
```
pcahyna: It appears that `type: partition` is not implemented.
```
TASK [storage : get required packages] *****************************************
task path: /home/pcahyna/linux-system-roles/storage/tasks/main-blivet.yml:88
fatal: [/home/pcahyna/linux-system-roles-testing/local/rhel-guest-image-7.7-200.x86_64.qcow2]: FAILED! => {"actions": [], "changed": false, "leaves": [], "mounts": [], "msg": "Pool 'foo' has unknown type 'partition'", "packages": [], "pools": [], "volumes": []}
```

tests/tests_disk_errors.yml
pcahyna: Moreover, even if I delete `type: partition` and add one volume in an effort to make it work, blivet refuses to create a VG on a disk which already contains a filesystem, even with safe mode off, so the test still fails:
```
TASK [storage : manage the pools and volumes to match the specified state] *****
task path: /home/pcahyna/linux-system-roles/storage/tasks/main-blivet.yml:102
fatal: [/home/pcahyna/linux-system-roles-testing/local/rhel-guest-image-7.7-200.x86_64.qcow2]: FAILED! => {"actions": [], "changed": false, "leaves": [], "mounts": [], "msg": "Failed to commit changes to disk", "packages": ["xfsprogs", "lvm2"], "pools": [], "volumes": []}
```

pcahyna:
@dwlehman the last changes are definitely an improvement. Changing the filesystem type does not destroy data anymore. Here is the error message:
```
TASK [storage : get required packages] *****************************************
task path: /home/pcahyna/linux-system-roles/storage/tasks/main-blivet.yml:88
fatal: [/home/pcahyna/linux-system-roles-testing/local/rhel-guest-image-7.7-200.x86_64.qcow2]: FAILED! => {"actions": [], "changed": false, "leaves": [], "mounts": [], "msg": "cannot remove existing formatting on volume 'test1' in safe mode", "packages": [], "pools": [], "volumes": []}
```
The assert
```yaml
- name: Verify the output
  assert:
    that: "blivet_output.failed and
           blivet_output.msg|regex_search('cannot remove existing formatting on volume.*in safe mode') and
              not blivet_output.changed"
    msg: "Unexpected behavior w/ existing data on specified disks"
```
still fails though, because `blivet_output` does not indicate failure (despite the actual failure):
```json
    "blivet_output": {
        "actions": [
            {
                "action": "create format",
                "device": "/dev/vdc",
                "fs_type": "ext4"
            }
        ],
        "changed": true,
        "failed": false,
        "leaves": [
            "/dev/sr0",
            "/dev/vda1",
            "/dev/vdb",
            "/dev/vdc"
        ],
        "mounts": [
            {
                "dump": 0,
                "fstype": "ext4",
                "opts": "defaults",
                "passno": 0,
                "path": "/opt/test1",
                "src": "/dev/vdc",
                "state": "mounted"
            }
        ],
        "packages": [
            "xfsprogs",
            "e2fsprogs"
        ],
        "pools": [],
        "volumes": [
            {
                "_device": "/dev/vdc",
                "_mount_id": "/dev/vdc",
                "disks": [
                    "vdc"
                ],
                "fs_create_options": "",
                "fs_label": "",
                "fs_overwrite_existing": true,
                "fs_type": "ext4",
                "mount_check": 0,
                "mount_device_identifier": "uuid",
                "mount_options": "defaults",
                "mount_passno": 0,
                "mount_point": "/opt/test1",
                "name": "test1",
                "size": 0,
                "state": "present",
                "type": "disk"
            }
        ]
    }
```

pcahyna:
tests improvements are in PR #46

tests/tests_disk_errors.yml
pcahyna: A fix for the test to not use `partition` for now is in #46

tests/tests_disk_errors.yml
pcahyna: PR #46 has the change to create one volume, but this still fails as in the previous comment.

tests/tests_disk_errors.yml
pcahyna: A more thorough test is in #46

tests/tests_disk_errors.yml
pcahyna: > Moreover, even if I delete `type: partition` and add one volume in an effort to make it work, blivet refuses to create a VG on a disk which already contains a filesystem, even with safe mode off, so the test still fails:
> 
> ```
> TASK [storage : manage the pools and volumes to match the specified state] *****
> task path: /home/pcahyna/linux-system-roles/storage/tasks/main-blivet.yml:102
> fatal: [/home/pcahyna/linux-system-roles-testing/local/rhel-guest-image-7.7-200.x86_64.qcow2]: FAILED! => {"actions": [], "changed": false, "leaves": [], "mounts": [], "msg": "Failed to commit changes to disk", "packages": ["xfsprogs", "lvm2"], "pools": [], "volumes": []}
> ```

Sorry, I think it was because I forgot to specify `size` for the volume. The error message is very misleading though

dwlehman:
@pcahyna I'm done until Tuesday. It looks good to me except for #48 which I do not think will be fixed here anyway.

pcahyna:
@dwlehman, more seriously, CI now fails in all the centos7/rhel7 runs, and always in the same place, with a Python TypeError, so it is not a CI infrastructure problem.

pcahyna:
> @dwlehman, more seriously, CI now fails in all the centos7/rhel7 runs, and always in the same place, with a Python TypeError, so it is not a CI infrastructure problem.

Hopefully corrected in #51

tests/tests_disk_errors.yml
```diff
@@ -22,3 +30,76 @@
...
+            that: "{{ not blivet_output.failed }}"
+            msg: "failed to create partition pool over existing file system using default value of safe_mode"
+
+        - name: Clean up
+          include_role:
+            name: storage
+          vars:
+            storage_volumes:
+              - name: test1
+                type: disk
```
pcahyna: @dwlehman  Why do we have `type: disk` in cleanup when we had `type: partition` in the previous task? Cleanup does not seem to happen properly.
Just replacing `disk` by `partition` does not work either though: 
... `Traceback (most recent call last):\r\n  File \"/root/.ansible/tmp/ansible-tmp-1571753120.2970297-255844982294535/AnsiballZ_blivet.py\", line 114, in <module>\r\n    _ansiballz_main()\r\n  File \"/root/.ansible/tmp/ansible-tmp-1571753120.2970297-255844982294535/AnsiballZ_blivet.py\", line 106, in _ansiballz_main\r\n    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)\r\n  File \"/root/.ansible/tmp/ansible-tmp-1571753120.2970297-255844982294535/AnsiballZ_blivet.py\", line 49, in invoke_module\r\n    imp.load_module('__main__', mod, module, MOD_DESC)\r\n  File \"/tmp/ansible_blivet_payload_ADJzqt/__main__.py\", line 784, in <module>\r\n  File \"/tmp/ansible_blivet_payload_ADJzqt/__main__.py\", line 781, in main\r\n  File \"/tmp/ansible_blivet_payload_ADJzqt/__main__.py\", line 747, in run_module\r\n  File \"/tmp/ansible_blivet_payload_ADJzqt/__main__.py\", line 568, in manage_volume\r\n  File \"/tmp/ansible_blivet_payload_ADJzqt/__main__.py\", line 230, in manage\r\n  File \"/tmp/ansible_blivet_payload_ADJzqt/__main__.py\", line 149, in _look_up_device\r\n  File \"/tmp/ansible_blivet_payload_ADJzqt/__main__.py\", line 283, in _get_device_id\r\nAttributeError: 'NoneType' object has no attribute '_disks'\r\n", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", "rc": 1`

tests/tests_disk_errors.yml
```diff
@@ -22,3 +30,76 @@
...
+          assert:
+            that: "{{ not blivet_output.failed }}"
+            msg: "failed to create partition pool over existing file system using default value of safe_mode"
+
+        - name: Clean up
+          include_role:
+            name: storage
+          vars:
+            storage_volumes:
+              - name: test1
```
pcahyna: @dwlehman Should be `foo`, no? Not `test1`

tests/tests_disk_errors.yml
pcahyna: Ah, maybe i have to change volumes to pools, partition is a pool type, not a volume type

tests/tests_disk_errors.yml
```diff
@@ -22,3 +30,76 @@
...
+                type: disk
+                disks: "{{ unused_disks }}"
+                present: false
```
pcahyna: Is there a `present` option? I would suppose that this would be handled via `state` (`present` or `absent`).

pcahyna: This commit https://github.com/linux-system-roles/storage/pull/54/commits/da350a9816351be09d010bf21ade80596a2317db in PR #54 hopefully fixes the cleanup issues.

tests/tests_disk_errors.yml
dwlehman: It is both a pool type and a volume type, much like `lvm`.

pcahyna: @dwlehman but
```yaml
          vars:
            storage_volumes:
              - name: test1
                type: partition
```
did not work, see the error above, I had to change to the version that is there now:
```yaml
      vars:
        storage_pools:
          - name: foo
            type: partition
            disks: "{{ unused_disks }}"
            state: absent
```

tasks/main-blivet.yml
```diff
@@ -94,6 +103,30 @@
...
+
+    - name: fail if mounts changed
+      fail:
+        msg: "get required packages changed mounts. Changed status is
+      {{ package_info.changed }}"
+  when:
+    - __storage_mounts_before_packages.ansible_facts.ansible_mounts |
+      count !=
+      __storage_mounts_after_packages.ansible_facts.ansible_mounts | count
+
```
pcahyna: @dwlehman shouldn't we make it a separate test rather than asserting every time the role is used? This commit was intended to quickly illustrate the problem, but maybe it is not ideal to keep it in the long term.

tests/tests_disk_errors.yml
dwlehman: Sorry, I didn't look closely enough at the context. There is indeed a 'partition' volume type but, just like the 'lvm' volume type, such volumes must be defined within a pool of the same type.

tasks/main-blivet.yml
dwlehman: Weird, I tried to leave that one out for that reason. Do you know which commit it came with? Maybe one of the merge commits?

pcahyna: Should be this:
Merge remote-tracking branch 'pcahyna/mount-assert' into non-destructive
57dd5ccf5f547c460d11ba2ab9434130e2a4a1cb

pcahyna:
well this change of class is on top of changes in this PR, it is not independent. If we want to simplify this PR, the change should be rather squashed with the one that introduces the problem. I spent about a week testing changes in this PR, so I believe it is OK.

dwlehman:
@tyll is this better now that it's been squashed to seven commits?

library/blivet.py
```diff
@@ -255,6 +262,19 @@ def _get_device_id(self):
     def _type_check(self):
         return self._device.is_disk
 
+    def _look_up_device(self):
+        super(BlivetDiskVolume, self)._look_up_device()
+        if not self._get_device_id():
+            # FAIL: no disks specified for volume
+            raise BlivetAnsibleError("no disks specified for volume '%s'" % self._volume['name'])  # sure about this one?
```
tyll: The comment here should be removed or the issue should be properly tracked in a ticket/addressed.

library/blivet.py
```diff
@@ -255,6 +262,19 @@ def _get_device_id(self):
...
+        elif not isinstance(self._volume['disks'], list):
+            raise BlivetAnsibleError("volume disks must be specified as a list")
+
+        if self._device is None:
+            # FAIL: failed to find the disk
```
tyll: The comment does not add extra information and should be removed. The next line says the same.

library/blivet.py
```diff
@@ -255,6 +262,19 @@ def _get_device_id(self):
...
+            raise BlivetAnsibleError("unable to resolve disk specified for volume '%s' (%s)" % (self._volume['name'], self._volume['disks']))
```
tyll: Does this error meant that no disks from `_volume["disks"]` contains the specified volume `_volume['name']`?
Without the code context the error message is hard to understand IMHO.

library/blivet.py
```diff
@@ -428,8 +448,11 @@ def _create_members(self):
         """ Schedule actions as needed to ensure pool member devices exist. """
         members = list()
         for disk in self._disks:
-            if not disk.isleaf:
-                self._blivet.devicetree.recursive_remove(disk)
+            if not disk.isleaf or disk.format.type is not None:
+                if not safe_mode:
```
tyll: The negation can be avoided by switching the statements to simplify the code:

```
if safe_mode:
    raise ...
else:
    self._blivet...
```

library/blivet.py
```diff
@@ -490,7 +513,10 @@ def _look_up_device(self):
     def _create(self):
         if self._device.format.type != "disklabel" or \
            self._device.format.label_type != disklabel_type:
-            self._blivet.devicetree.recursive_remove(self._device, remove_device=False)
+            if not safe_mode:
```
tyll: The negation can be avoided here as well.

library/blivet.py
```diff
@@ -660,6 +686,7 @@ def run_module():
         volumes=dict(type='list'),
         packages_only=dict(type='bool', required=False, default=False),
         disklabel_type=dict(type='str', required=False, default=None),
+        safe_mode=dict(type='bool', required=False, default=False),
```
tyll: Shouldn't the default be `True` here for safety? The documentation implies that `True` is the default and it is implemented via the role variables only.

tests/tests_disk_errors.yml
```diff
@@ -54,13 +80,40 @@
...
           assert:
-            that: "{{ blivet_output.failed and
-                      blivet_output.msg|regex_search('cannot remove existing formatting on volume.*in safe mode') and
-                      not blivet_output.changed }}"
-            msg: "Unexpected behavior w/ existing data on specified disks"
+            that:
+              - ansible_failed_task.name != 'UNREACH'
+            msg: "Role has not failed when it should have"
+          vars:
+            # Ugh! needed to expand ansible_failed_task
```
tyll: This comment seems to indicate that there is still something that needs to be done but to me it is unclear what is meant.

library/blivet.py
```diff
@@ -120,6 +120,7 @@
 use_partitions = None  # create partitions on pool backing device disks?
 disklabel_type = None  # user-specified disklabel type
 safe_mode = None       # do not remove any existing devices or formatting
+packages_only = None   # only set things up enough to get a list of required packages
```
tyll: This commit does not seem to fit in this PR and it seems that the added functionality is not used or tested. Not sure what it is doing here.

tyll:
> @tyll is this better now that it's been squashed to seven commits?

AFAICS, it seems that the first commit is the primary reason for this PR. IMHO the other commits that do not really relate to the change and do not need any more discussion (like using the new-style classes) should be moved to its own PR and already be merged.

Squashing the test fixes does not help to properly review them because now the commit messages do not relate to the changes.

Too much change at once, especially when it also requires discussion, makes this too confusing or too hard to keep track of everything.

tests/tests_lvm_errors.yml
```diff
@@ -149,3 +149,87 @@
...
+              - name: foo
+                disks: "{{ unused_disks }}"
+                type: lvm
+
+        - name: Verify the output
+          assert:
+            that: "{{ blivet_output.failed and
+                      blivet_output.msg|regex_search('cannot remove existing formatting and/or devices on disk.*in safe mode') and
+                      not blivet_output.changed }}"
+            msg: "Unexpected behavior w/ existing data on specified disks"
```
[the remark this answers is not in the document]
dwlehman: These are only the output of integration tests, but okay.

library/blivet.py
```diff
@@ -31,6 +31,9 @@
     disklabel_type:
         description:
             - disklabel type string (eg: 'gpt') to use, overriding the built-in logic in blivet
+    safe_mode:
+        description:
+            - boolean indicating that we should fail rather than implicitly/automatically removing devices or formatting
```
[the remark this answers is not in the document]
dwlehman: This requirement is arbitrary at best and silly at worst. How did you arrive at 88 chars as a max? It was once 65 or 68, then it was 70-something, then 80, now 88? Most displays are 1200px or wider, so I don't understand the need to limit line lengths like we did when there was a fixed console width, but I also don't want to spend too much time arguing about it.

library/blivet.py
dwlehman: Yes, it should. Good catch.

tests/tests_disk_errors.yml
dwlehman: @pcahyna can you answer this one?

library/blivet.py
dwlehman: Sure, I can move this out.

library/blivet.py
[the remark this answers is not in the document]
tyll: The 88 character suggestion comes from the black formatter https://github.com/psf/black#line-length - to me the arguments make sense. I do not want to argue about it, I just use black.

Regarding the display size - I have a 4096 px display, however as I wrote GitHub does not use all the space to show your long line but requires vertical scrolling. This makes it painful to read these lines. Also it is best-practice in the analog world to wrap lines at a certain point to make them easier to read, see newspapers and journals, because of the width of the human vision.

Now you have this information it is up to you if you want to make it unpleasant to look at your patches or make it a pleasant experience. I do not intend to force you to do anything.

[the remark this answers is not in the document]
dwlehman: For me, the point of being unpleasantly (or oddly) long is around 120 characters.

I do not have any interest in changing my code (or code comments) to make the github webui experience better, but I still take your point about there being a limit somewhere.

tests/tests_disk_errors.yml
pcahyna: It is a workaround for a silly behavior in Ansible. The problem is, `ansible_failed_task.name` references the `storage_provider` variable which is not defined anymore at this point. But variables are evaluated lazily, so by evaluating `ansible_failed_task.name` here one crashes Ansible with an undefined variable reference. It should be done better, but I don't know how. (True, the comment should describe the problem better so it is not lost.)

pcahyna:
> > @tyll is this better now that it's been squashed to seven commits?
> 
> AFAICS, it seems that the first commit is the primary reason for this PR. IMHO the other commits that do not really relate to the change and do not need any more discussion (like using the new-style classes) should be moved to its own PR and already be merged.

You are probably missing that without the new-style class fix, the code in the first commit simply does not work (https://github.com/linux-system-roles/storage/pull/43/commits/399aee20543d56c63ad076da456f4b3510fb7b22#diff-bdb43621871ab82b00fbc3e2dcc1c681R265 ). I could remove the other new-stylle class changes and keep just the one required to make the PR work though.

Without my other changes tests do not test anything and it is wrong to merge a pull request with falsely passing tests IMO.

tests/tests_disk_errors.yml
pcahyna: here are references to Ansible bug reports: https://github.com/ansible/ansible/issues/49942 https://github.com/ansible/ansible/issues/57399

pcahyna: @tyll  For the record, #88 avoids the problem. https://github.com/linux-system-roles/network/pull/200 shows a different approach.

PR #46, opened 2019-10-15
Many fixes for the test in PR #43

pcahyna:
To be applied on top of it.

Note that the tests still fail despite the fixes. The original version passed by accident - ignore_errors was set on all relevant tasks.

### DIFF
--- a/tests/tests_disk_errors.yml
+++ b/tests/tests_disk_errors.yml
@@ -5,6 +5,14 @@
     mount_location: '/opt/test1'
 
   tasks:
+    - include_role:
+        name: storage
+
+    - include_tasks: get_unused_disk.yml
+      vars:
+        min_size: "10g"
+        max_return: 1
+
     - name: Verify that the play fails with the expected error message
       block:
         - name: Create a disk volume mounted at "{{ mount_location }}"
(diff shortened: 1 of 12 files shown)


tests/tests_disk_errors.yml
```diff
@@ -124,36 +124,61 @@
...
+      include_role:
+        name: storage
+      vars:
+        storage_safe_mode: false
+        storage_pools:
+          - name: foo
+            disks: "{{ unused_disks }}"
+#            type: partition
+            volumes:
+              - name: bar
```
dwlehman: I think we already have a test that does this for LVM, no?

tests/tests_disk_errors.yml
```diff
@@ -122,7 +122,9 @@
             storage_pools:
               - name: foo
                 disks: "{{ unused_disks }}"
-                type: partition
+#                type: partition
+                volumes:
+                  - name: bar
```
dwlehman: It's implemented, just an incorrect key in the class lookup dict. It's fixed in #43.

tests/tests_disk_errors.yml
pcahyna: This is very well possible. The test did not work as it was (with `partition`) so I made a small change to improve it.

tests/tests_disk_errors.yml
pcahyna: I will revert this part then

dwlehman:
I did a force-push to update 'Fix logic of safe mode guard on reformat.'

Picking that up should help at least with the failure CI is hitting now.

pcahyna:
@dwlehman don't you want to merge this PR into your #43 ?

pcahyna:
> I did a force-push to update 'Fix logic of safe mode guard on reformat.'

Where did you push it? I don't see any new commit.

pcahyna:
Sorry, I see it now.

dwlehman:
Yeah, sorry about that. I didn't realize you were following my work so closely at the time. I won't be doing any more force-pushes to avoid the confusion they can cause.

pcahyna:
No problem, just please avoid touching the same test. But since you need a test for your work, I think including this PR in your PR would be better. By the way the rebase did not help. It still fails (unmounts the filesystem and therefore the test believes it lost data).

pcahyna:
By the way, I created a separate assert for the problem that is now causing the test to fail: #47

PR #47, opened 2019-10-16
Assert that "get required packages" has not changed mounts

pcahyna:
It seems that the "get required packages" task (invocation of blivet with
packages_only: true) silently (without even reporting changed: true) changes
mounts. Add an assert for that. (It would be better to make it a separate unit
test.)

### DIFF
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -85,6 +85,15 @@
 - debug:
     var: _storage_volumes
 
+- name: load mount facts
+  setup:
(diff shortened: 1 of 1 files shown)


richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

richm:
do we still need this?

richm:
[citest pending]

richm:
closing due to inactivity

PR #51, opened 2019-10-21
Make all classes new-style (derive from object)

pcahyna:
Otherwise super() fails. Not necessary with Python 3, but necessary with Python 2 (where classic classes are the default)


dwlehman:
Also added to #43.

pcahyna:
remaining missing change now committed in #72

PR #55, opened 2019-10-21
Clone of #43 to test LGTM

pcahyna:


pcahyna:
@tyll lgtm did not work for this particular bug. Our CI catches failures with Python2 (fixed in #51) that lgtm could have catched as well.

tyll:
> @tyll lgtm did not work for this particular bug. Our CI catches failures with Python2 (fixed in #51) that lgtm could have catched as well.

It seems to check only for Python 3 bugs for the storage role, according to https://lgtm.com/help/lgtm/analysis-faqs#how-python-version-identified it supports only one Python version and the logs for the storage role check show that it selected Python 3 here: https://lgtm.com/projects/g/linux-system-roles/storage/logs/languages/lang:python

Python 2 could be specifically selected:
https://lgtm.com/help/lgtm/python-extraction

pcahyna:
@tyll how is then possible that it found a very similar problem in the network role, here: https://github.com/linux-system-roles/network/pull/131/files#diff-d32c4e7fad7beec907c9d463e17228d4R1376 ? Is it using Python 2 for network? If so, it should probably be configured to use Python 2 for all the roles.

tyll:
#62 contains a config file to select python 2, you can include the commit here to test if it properly identifies the expected bugs.

pcahyna:
@tyll it works, thanks, the relevant alert is "1 for 'super' in old style class". What is surprising is that despite the correct alerts, the test still shows as passed (green).

Issue #49, opened 2019-10-17
`eq` or `equalto` Jinja operators are not supported in Jinja 2.7.2 that is in RHEL7


pcahyna:
Even `eq` is not supported in Jinja 2.7.2 that is in RHEL7 . We need to use `==`. @yontalcar

_Originally posted by @pcahyna in https://github.com/linux-system-roles/storage/pull/26#issuecomment-543200427_

yontalcar:
Problem here is that it is used as a test in `selectattr` filter:
https://github.com/linux-system-roles/storage/blob/master/tests/test-verify-volume-mount.yml#L4

pcahyna:
@yontalcar then it needs to be rewritten - https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#json-query-filter perhaps?

pcahyna:
Fixed in #56

PR #56, opened 2019-10-22
Replace selectattr filter with json_query filter

yontalcar:
Some tests (like comparison) are missing in older Jinja. That makes
selectattr filter unusable on RHEL 7.

### DIFF
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -51,6 +51,8 @@
 
 - name: Apply defaults to pools and volumes [4/6]
   set_fact:
+    # json_query(...) used instead of "|selectattr('pool', 'equalto', item.name)|list"
+    # as that expression wouldn't work with Jinja versions <2.8
     _storage_vols_no_defaults_by_pool: "{{ _storage_vols_no_defaults_by_pool|default({})|
                                         combine({item.name: _storage_vols_w_defaults|json_query('[?pool==`\"{}\"`]'.format(item.name))}) }}"
   loop: "{{ _storage_pools }}"
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -52,7 +52,7 @@
 - name: Apply defaults to pools and volumes [4/6]
   set_fact:
     _storage_vols_no_defaults_by_pool: "{{ _storage_vols_no_defaults_by_pool|default({})|
-                                        combine({item.name: _storage_vols_w_defaults|selectattr('pool', 'equalto', item.name)|list}) }}"
+                                        combine({item.name: _storage_vols_w_defaults|json_query('[?pool==`\"{}\"`]'.format(item.name))}) }}"
   loop: "{{ _storage_pools }}"
   when: storage_pools is defined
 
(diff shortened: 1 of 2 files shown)


pcahyna:
fixes #49 (hopefully)

tasks/main-blivet.yml
```diff
@@ -52,7 +52,7 @@
 - name: Apply defaults to pools and volumes [4/6]
   set_fact:
     _storage_vols_no_defaults_by_pool: "{{ _storage_vols_no_defaults_by_pool|default({})|
-                                        combine({item.name: _storage_vols_w_defaults|selectattr('pool', 'equalto', item.name)|list}) }}"
+                                        combine({item.name: _storage_vols_w_defaults|json_query('[?pool==`{}`]'.format(item.name))}) }}"
```
pcahyna: what is the meaning of the backticks (`` `{}` ``) ?

yontalcar: AIUI, backticks are just another quotation marks.
`{}` is a placeholder for format

pcahyna: > AIUI, backticks are just another quotation marks.

Is it documented anywhere? I checked Jinja docs and they do not mention anything about backticks.

yontalcar: Hmm… after reading specs, backticks are used to enclose JSON value… I should use either `` `"string"` `` or `'string'`

pcahyna: @yontalcar  Can you link the specs you are using please?

yontalcar: it's JMESPath thing: http://jmespath.org/specification.html

yontalcar:
[citest bad]

pcahyna:
@dwlehman if there are no objections, I would like to merge it, because otherwise we have a problem with the soreage role on RHEL 7.

tasks/main-blivet.yml
```diff
@@ -52,7 +52,7 @@
 - name: Apply defaults to pools and volumes [4/6]
   set_fact:
     _storage_vols_no_defaults_by_pool: "{{ _storage_vols_no_defaults_by_pool|default({})|
-                                        combine({item.name: _storage_vols_w_defaults|selectattr('pool', 'equalto', item.name)|list}) }}"
+                                        combine({item.name: _storage_vols_w_defaults|json_query('[?pool==`\"{}\"`]'.format(item.name))}) }}"
```
pcahyna: I suggest to leave the original selectattr filter in a comment above, since it is probably more readable.

tests/test-verify-volume-mount.yml
```diff
@@ -1,8 +1,8 @@
 ---
 - name: Set some facts
   set_fact:
-    storage_test_mount_device_matches: "{{ ansible_mounts|selectattr('device', 'equalto', storage_test_volume._device)|list }}"
-    storage_test_mount_point_matches: "{{ ansible_mounts|selectattr('mount', 'equalto', storage_test_volume.mount_point)|list }}"
+    storage_test_mount_device_matches: "{{ ansible_mounts|json_query('[?device==`\"{}\"`]'.format(storage_test_volume._device))}}"
+    storage_test_mount_point_matches: "{{ ansible_mounts|json_query('[?mount==`\"{}\"`]'.format(storage_test_volume.mount_point))}}"
```
pcahyna: I suggest to leave the original selectattr filter in a comment above, since it is probably more readable.

Issue #48, opened 2019-10-16
If an option is omitted on existing volume, it gets reset to the default. 


pcahyna:
Omitted `mount_point`: the volume gets unmounted.

Type switching: if I omit `fs_type:`, the volume gets reformatted with the default type (xfs). Wouldn't it be more reasonable to keep the existing filesystem? (note that with `storage_safe_mode` introduced in #43, any destructive operation is prevented - the role errors out. Still, there is the question whether even attempting to change the type to default is reasonable.)

_Originally posted by @pcahyna in https://github.com/linux-system-roles/storage/pull/43#issuecomment-540740180_

pcahyna:
Quoting @dwlehman 's reply:

> I'm not sure about that. I can see a case for either behavior. I have a somewhat difficult time imagining a user omitting fs_type and having a specific expectation about the result. Probably some people will complain whichever way we choose to interpret it.

Quoting my reply:

> The scenario I am imagining is that someone will take a storage configuration created before by some other tools (anaconda, manually...), already used for storing data, and they will start using the storage role to manage it - change its properties, most likely size. It can be unexpected to see the the storage role reformatting your volumes (or at least trying to remove them, with the changes in #43 ) and recreating them just because you did specify only the size (that you want to change) and not the filesystem type (or any other attribute that should be left unchanged).

dwlehman:
There has to be a way to specify that a given volume not be mounted, obviously.

We may need more sophisticated default handling if we're going to need to differentiate between "the default" and "whatever is already there". By the way, the latter is invalid for a non-existent volume.

pcahyna:
> There has to be a way to specify that a given volume not be mounted, obviously.

Of course. ZFS uses mountpoint=none for this. I think this already works in the role.

> We may need more sophisticated default handling if we're going to need to differentiate between "the default" and "whatever is already there". By the way, the latter is invalid for a non-existent volume.

Sure. The latter could be defined as "default for new volumes, whatever is already there for existing volumes". By the way, how exactly does `size` work now? Does the role change the size of a volume to a default if not specified? Or there is no default?

dwlehman:
`size` is required, although I think you found that we aren't explicitly checking that.

robled:
> We may need more sophisticated default handling if we're going to need to differentiate between "the default" and "whatever is already there". By the way, the latter is invalid for a non-existent volume.

[...]

PR #59, opened 2019-10-24
Missing parameters

dwlehman:
This changes the behavior of the role such that omitted parameters do not change settings on existing pools and volumes. Omitted parameters on new pools and volumes will be given the default value, while the same on existing pools and volumes will leave whatever the current setting is intact.

### DIFF
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -9,90 +9,20 @@
     name: "{{ blivet_package_list }}"
     state: present
 
-##
-## This is all to fill in missing defaults in the list of pools (and their volumes).
-##
-## The pools are easy, as are the standalone volumes. The pool volumes are trickier,
-## perhaps because I am new to ansible/yaml/jinja.
-##
-- name: initialize internal facts
-  set_fact:
-    _storage_pools: []  # list of pools w/ defaults applied as necessary
-    _storage_volumes: []  # list of standalone volumes w/ defaults applied as necessary
-    _storage_vol_defaults: []  # list w/ volume default dict for each pool volume
-    _storage_vols_no_defaults: []  # combined list of pool volumes w/o defaults applied
-    _storage_vols_w_defaults: []  # combined list of volumes w/ defaults applied
-    _storage_vol_pools: []  # combined list of pool name for each volume
-    _storage_vols_no_defaults_by_pool: {}  # dict w/ pool name keys and pool volume list values
-
-- name: Apply defaults to pools and volumes [1/6]
-  set_fact:
-    _storage_pools: "{{ _storage_pools|default([]) }} + [ {{ storage_pool_defaults|combine(pool) }} ]"
-  loop: "{{ storage_pools|default([]) }}"
-  loop_control:
-    loop_var: pool
-  when: storage_pools is defined
-
-- name: Apply defaults to pools and volumes [2/6]
-  set_fact:
-    _storage_vols_no_defaults: "{{ _storage_vols_no_defaults|default([]) }} + [{{ item.1 }}]"
-    _storage_vol_defaults: "{{ _storage_vol_defaults|default([]) }} + [{{ storage_volume_defaults }}]"
-    _storage_vol_pools: "{{ _storage_vol_pools|default([]) }} + ['{{ item.0.name }}']"
-  loop: "{{ _storage_pools|subelements('volumes', skip_missing=true) }}"
-  when: storage_pools is defined
-
-- name: Apply defaults to pools and volumes [3/6]
-  set_fact:
-    _storage_vols_w_defaults: "{{ _storage_vols_w_defaults|default([]) }} + [ {{ item.1|combine(item.0, {'pool': _storage_vol_pools[idx]}) }} ]"
-  loop: "{{ _storage_vols_no_defaults|zip(_storage_vol_defaults)|list }}"
-  loop_control:
-    index_var: idx
-  when: storage_pools is defined
-
-- name: Apply defaults to pools and volumes [4/6]
-  set_fact:
-    # json_query(...) used instead of "|selectattr('pool', 'equalto', item.name)|list"
-    # as that expression wouldn't work with Jinja versions <2.8
-    _storage_vols_no_defaults_by_pool: "{{ _storage_vols_no_defaults_by_pool|default({})|
-                                        combine({item.name: _storage_vols_w_defaults|json_query('[?pool==`\"{}\"`]'.format(item.name))}) }}"
-  loop: "{{ _storage_pools }}"
-  when: storage_pools is defined
-
-- name: Apply defaults to pools and volumes [5/6]
-  set_fact:
-    _storage_pools: "{{ _storage_pools[:idx] }} +
-                     [ {{ pool|combine({'volumes': _storage_vols_no_defaults_by_pool[pool.name]}) }} ] +
-                     {{ _storage_pools[idx+1:] }}"
-  loop: "{{ _storage_pools }}"
-  loop_control:
-    loop_var: pool
-    index_var: idx
-  when: storage_pools is defined
-
-- name: Apply defaults to pools and volumes [6/6]
-  set_fact:
-    _storage_volumes: "{{ _storage_volumes|default([]) }} + [ {{ storage_volume_defaults|combine(volume) }} ]"
-  loop: "{{ storage_volumes|default([]) }}"
-  loop_control:
-    loop_var: volume
-  when: storage_volumes is defined
-
-##
-## End of absurdly long process to inject defaults into user-specified pools and volumes lists.
-##
+- debug:
+    var: storage_pools
 
 - debug:
-    var: _storage_pools
-
-- debug:
-    var: _storage_volumes
+    var: storage_volumes
 
 - name: get required packages
   blivet:
-    pools: "{{ _storage_pools }}"
-    volumes: "{{ _storage_volumes }}"
+    pools: "{{ storage_pools|default([]) }}"
+    volumes: "{{ storage_volumes|default([]) }}"
     use_partitions: "{{ storage_use_partitions }}"
     disklabel_type: "{{ storage_disklabel_type }}"
+    pool_defaults: "{{ storage_pool_defaults }}"
+    volume_defaults: "{{ storage_volume_defaults }}"
     packages_only: true
   register: package_info
 
@@ -103,10 +33,12 @@
 
 - name: manage the pools and volumes to match the specified state
   blivet:
-    pools: "{{ _storage_pools }}"
-    volumes: "{{ _storage_volumes }}"
+    pools: "{{ storage_pools|default([]) }}"
+    volumes: "{{ storage_volumes|default([]) }}"
     use_partitions: "{{ storage_use_partitions }}"
     disklabel_type: "{{ storage_disklabel_type }}"
+    pool_defaults: "{{ storage_pool_defaults }}"
+    volume_defaults: "{{ storage_volume_defaults }}"
     safe_mode: "{{ storage_safe_mode }}"
     diskvolume_mkfs_option_map: "{{ __storage_blivet_diskvolume_mkfs_option_map|d(omit) }}"
   register: blivet_output
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -31,17 +31,40 @@
     name: "{{ package_info.packages }}"
     state: present
 
-- name: manage the pools and volumes to match the specified state
-  blivet:
-    pools: "{{ storage_pools|default([]) }}"
-    volumes: "{{ storage_volumes|default([]) }}"
-    use_partitions: "{{ storage_use_partitions }}"
-    disklabel_type: "{{ storage_disklabel_type }}"
-    pool_defaults: "{{ storage_pool_defaults }}"
-    volume_defaults: "{{ storage_volume_defaults }}"
-    safe_mode: "{{ storage_safe_mode }}"
-    diskvolume_mkfs_option_map: "{{ __storage_blivet_diskvolume_mkfs_option_map|d(omit) }}"
-  register: blivet_output
+- service_facts:
+
+- set_fact:
+    # For an explanation of the to_json|from_json silliness, see
+    # https://github.com/ansible-collections/community.general/issues/320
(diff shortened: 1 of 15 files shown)


dwlehman:
I see there are CI failures. I'll take a look on Monday.

pcahyna:
@dwlehman thanks, but I would prefer to wait with this for further user feedback, as the discussion in #48 shows, it is not clear to what extent it is desirable or not.

dwlehman:
I only looked over it briefly, but from what I gathered everyone there is in agreement that we should not change settings for existing volumes because of omitted parameters.

dwlehman:
I just rebased this and added the commit allowing lookup of an existing pool by name only. I ran one random integration test and it passed, so I'm waiting to see what CI turns up.

dwlehman:
To Do:
- type-specific parameters (lvm, md, luks)
- docs

tests/run_blivet.yml
```diff
@@ -1,7 +1,13 @@
 ---
+- include_vars:
+    dir: roles/storage/defaults
+    extensions: ['yml']
```
richm, line 4: What's this for?  I guess this is to have access to the default role variables in the test, before the role is included?  Can you do `include_vars:  name: roles/storage/defaults/main.yml`?

dwlehman: That is correct and sure -- I can explicitly include the main.yml. That's what I wanted to do anyway.

tests/run_blivet.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - include_vars:
-    name: roles/storage/defaults/main.yml
+    file: roles/storage/defaults/main.yml
```
richm, line 3: lgtm

richm: please use `file: "{{ playbook_dir }}/roles/linux-system-roles.storage/defaults/main.yml"`

tests/tests_existing_lvm_pool.yml
```diff
@@ -0,0 +1,54 @@
...
+  become: true
+  vars:
+    mount_location: '/opt/test1'
+    volume_group_size: '5g'
+    volume_size: '4g'
+    pool_name: foo
+
+  tasks:
+    - include_role:
+        name: storage
```
richm, line 12: please use `name: linux-system-roles.storage`

tests/tests_existing_lvm_pool.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - include_tasks: get_unused_disk.yml
+      vars:
+        min_size: "{{ volume_group_size }}"
+        max_return: 1
+
+    - name: Create one LVM logical volume under one volume group
+      include_role:
+        name: storage
```
richm, line 21: please use `name: linux-system-roles.storage`

tests/tests_existing_lvm_pool.yml
```diff
@@ -0,0 +1,54 @@
...
+              disks: "{{ unused_disks }}"
+              volumes:
+                - name: test1
+                  size: "{{ volume_size }}"
+
+    - include_tasks: verify-role-results.yml
+
+    - name: Create another volume in the existing pool, identified only by name.
+      include_role:
+        name: storage
```
richm, line 34: please use `name: linux-system-roles.storage`

tests/tests_existing_lvm_pool.yml
```diff
@@ -0,0 +1,54 @@
...
+              - name: newvol
+                size: '2 GiB'
+                fs_type: ext4
+                fs_label: newvol
+
+    - include_tasks: verify-role-results.yml
+
+    - name: Clean up.
+      include_role:
+        name: storage
```
richm, line 48: please use `name: linux-system-roles.storage`

tests/tests_luks_pool.yml
```diff
@@ -109,10 +110,11 @@
 
     - include_tasks: verify-role-results.yml
 
-    - name: Add encryption to the volume
+    - name: Add both encryption and raid to the pool
       include_role:
         name: storage
```
richm, line 115: please use `name: linux-system-roles.storage`

tests/tests_luks_pool.yml
```diff
@@ -133,6 +136,7 @@
...
```
richm, line 137: please use `name: linux-system-roles.storage`

meta/main.yml
```diff
@@ -7,6 +7,6 @@ galaxy_info:
   min_ansible_version: 2.5
   platforms:
     - name: Fedora
-      versions: [ 31, 32 ]
+      versions: [ 31, 32, 33 ]
```
richm, line 10: @jharuda can this be now changed to just `versions: all`?

richm:
[citest]

dwlehman:
_Note to self, since I'm taking Friday off_:

There are two persistent CI failures that I have been unable to reproduce locally, both of which appear to be triggered by the CI infrastructure's slowness:

1. udevd is failing to pick up the label after the `mkfs; e2label` sequence, which might be resolved using `flock` in blivet (`tests_existing_pool.yml` on f32)
2. some devices/filesystems are active on disks, causing a parted commit failure (`tests_luks_pool.yml` on rhel-7)

I can investigate blivet mitigation for failure 1, but I'll have to figure out how to examine the blivet log from the actual ci run in order to diagnose failure 2.

dwlehman:
[citest]

dwlehman:
[citest bad]

dwlehman:
[citest bad]

dwlehman:
[citest bad]

dwlehman:
[citest bad]

dwlehman:
[citest bad]

dwlehman:
It's worth noting that this is a pretty big behavioral change, and probably merits a major version bump. @pcahyna @richm thoughts on that?

richm:
> It's worth noting that this is a pretty big behavioral change, and probably merits a major version bump. @pcahyna @richm thoughts on that?

Are there changes which will break the API e.g. if a customer has playbooks that work with the existing role, will those playbooks continue to work the exact same way if they update the role?
If there are breaking changes, we will need to bump the major version - and also figure out a way to communicate this to users of the role so that they can modify their playbooks and/or inventories, and justify such a change.

tasks/main-blivet.yml
```diff
@@ -101,15 +31,39 @@
...
-- name: manage the pools and volumes to match the specified state
-  blivet:
-    pools: "{{ _storage_pools }}"
-    volumes: "{{ _storage_volumes }}"
-    use_partitions: "{{ storage_use_partitions }}"
-    disklabel_type: "{{ storage_disklabel_type }}"
-    safe_mode: "{{ storage_safe_mode }}"
-    diskvolume_mkfs_option_map: "{{ __storage_blivet_diskvolume_mkfs_option_map|d(omit) }}"
-  register: blivet_output
+- command: 'systemctl list-units --all --plain --full --no-legend -t service'
```
richm, line 34: Can you use `service_facts` instead?  https://docs.ansible.com/ansible/2.9/modules/service_facts_module.html#service-facts-module

tasks/main-blivet.yml
```diff
@@ -101,15 +31,39 @@
...
+      volumes: "{{ storage_volumes|default([]) }}"
+      use_partitions: "{{ storage_use_partitions }}"
+      disklabel_type: "{{ storage_disklabel_type }}"
+      pool_defaults: "{{ storage_pool_defaults }}"
+      volume_defaults: "{{ storage_volume_defaults }}"
+      safe_mode: "{{ storage_safe_mode }}"
+      diskvolume_mkfs_option_map: "{{ __storage_blivet_diskvolume_mkfs_option_map|d(omit) }}"
+    register: blivet_output
+  rescue:
+    - fail:
```
richm, line 60: If you want to check the result of the failure in tests, you'll have to "re-raise" the error - see https://richm.github.io/how-to-catch-and-reraise-errors-in-ansible

dwlehman:
> > It's worth noting that this is a pretty big behavioral change, and probably merits a major version bump. @pcahyna @richm thoughts on that?
> 
> Are there changes which will break the API e.g. if a customer has playbooks that work with the existing role, will those playbooks continue to work the exact same way if they update the role?

No, the behavior will change when they run the role against pre-existing pools or volumes with any parameters omitted. It seems clear that the new behavior is "better" (don't change anything unless explicitly told to), but it's definitely different.

> If there are breaking changes, we will need to bump the major version - and also figure out a way to communicate this to users of the role so that they can modify their playbooks and/or inventories, and justify such a change.

As I said above, the new behavior is less disruptive/destructive, but it's still different.

richm:
> As I said above, the new behavior is less disruptive/destructive, but it's still different.

Will a playbook run against the new role report any differences at all?
* run playbook against old role
* upgrade role
* run playbook again with new role and no other changes

Will the playbook run report `changed`?  That is, will the upgrade break idempotency (assuming the role is currently idempotent)?  If so, then this suggests to me a major version change.

dwlehman:
> > As I said above, the new behavior is less disruptive/destructive, but it's still different.
> 
> Will a playbook run against the new role report any differences at all?
> 
>     * run playbook against old role
> 
>     * upgrade role
> 
>     * run playbook again with new role and no other changes
> 
> 
> Will the playbook run report `changed`? That is, will the upgrade break idempotency (assuming the role is currently idempotent)? If so, then this suggests to me a major version change.

No, it should not report 'changed' in that case.

richm:
> No, it should not report 'changed' in that case.

Then IMO a major version change is not needed, only a minor version change.

tasks/main-blivet.yml
dwlehman: Changed to use `service_facts` in latest version. Bonus fun ensued related to `json_query`.

tasks/main-blivet.yml
dwlehman: Fixed in latest version, although tests were passing as it was.

richm: > Fixed in latest version, although tests were passing as it was.

Then you probably aren't trying to `rescue` and check the error in a calling context.

tasks/main-blivet.yml
```diff
@@ -31,17 +31,40 @@
...
-    volume_defaults: "{{ storage_volume_defaults }}"
-    safe_mode: "{{ storage_safe_mode }}"
-    diskvolume_mkfs_option_map: "{{ __storage_blivet_diskvolume_mkfs_option_map|d(omit) }}"
-  register: blivet_output
+- service_facts:
+
+- set_fact:
+    # For an explanation of the to_json|from_json silliness, see
+    # https://github.com/ansible-collections/community.general/issues/320
+    storage_cryptsetup_services: "{{ ansible_facts.services|to_json|from_json|json_query('*.name')|json_query('[?starts_with(@, `\"systemd-cryptsetup@\"`)]') }}"
```
richm, line 39: Here is a way to do it without json_query:
```
    storage_cryptsetup_services: "{{
      ansible_facts.services | dict2items | selectattr('key', 'match', '^systemd-cryptsetup@') |
      map(attribute='value') | map(attribute='name') | list
    }}"
```
however, on fedora 33, this gives me a value like this: `['systemd-cryptsetup@luks\\x2d7fa23e85\\x2d4729\\x2d43b2\\x2dcc7a\\x2d9608d8cdb732.service']`
not sure if that is the same result of using json_query

dwlehman: Someone came in a while back and switched all of our `selectattr` usage to `json_query` since `selectattr` doesn't work with Jinja versions < 2.8. You can see the comment, look at blame, etc. in `storage/tests/test-verify-volume-mount.yml`.

richm: > Someone came in a while back and switched all of our `selectattr` usage to `json_query` since `selectattr` doesn't work with Jinja versions < 2.8. You can see the comment, look at blame, etc. in `storage/tests/test-verify-volume-mount.yml`.

Mea culpa.  That was based on my incomplete understanding of the problems of `selectattr` and `json_query`.  The problem with jinja27 is not `selectattr` in general, it is that certain filters are not available - specifically, the `equalto`, `equal`, `==` filters to test if a value in an object is equal to some given value.

We have just completed converting all of the system roles to work with jinja27, and we figured out how to get around this limitation, by using the `match` filter instead of `equalto` or `==` - https://github.com/linux-system-roles/logging/commit/48d21efb71a36073a75914bdc3251dbb7fa63e77#diff-fe77ca1acb10f971c92988a8aed5bb5ae631443b1a20668629ccc846c608205aL11-L12

In the case of this PR, we have to use `match` anyway to look for services beginning with `systemd-cryptsetup@`.

In addition, I have verified the code above works with ansible 2.8/jinja2.7/centos7 controller host.

richm: We have CI tests for ansible 2.8/jinja2.7/centos7 controller host - the statuses like `*/ansible-2.8` - so you can know if you make a change that breaks in this case.

richm: However, all of that being said, if you prefer to keep the `json_query` implementation, that's fine with me.

dwlehman: I'm happy with it as it is now if you are. No need to fiddle with it any more.

dwlehman:
[citest bad]

richm:
Don't worry about the `(staging)` failures.  All of the required tests have passed.

richm:
The `rhel-x/ansible-2.9` fails because of this:
```
TASK [Read the /etc/crypttab file] *********************************************
task path: /tmp/tmpppms2nvo/tests/verify-role-results.yml:24
fatal: [/cache/rhel-x.qcow2]: FAILED! => {"changed": false, "cmd": ["cat", "/etc/crypttab"], "delta": "0:00:00.002482", "end": "2021-02-08 20:39:44.912470", "msg": "non-zero return code", "rc": 1, "start": "2021-02-08 20:39:44.909988", "stderr": "cat: /etc/crypttab: No such file or directory", "stderr_lines": ["cat: /etc/crypttab: No such file or directory"], "stdout": "", "stdout_lines": []}
```
rhel-x uses the internal rhel 8.4.0 snapshot 1 - which is quite old by now - we are in the process of upgrading it - is it possible that this test uses a component that had a bug in this older snapshot, or the test uses a component in rhel 8.4 that is in a later version of 8.4?

richm:
[citest bad]

richm (approved):
lgtm

pcahyna:
> rhel-x uses the internal rhel 8.4.0 snapshot 1 - which is quite old by now - we are in the process of upgrading it - is it possible that this test uses a component that had a bug in this older snapshot, or the test uses a component in rhel 8.4 that is in a later version of 8.4?

No, crypttab is still missing when testing against the latest nightly snapshot.

PR #53, opened 2019-10-21
Test fixes for lvm

pcahyna:
Analogy of PR #46 but for the LVM test that I missed initially.

PR #64, opened 2019-11-05
Fixes for bugs found during testing of 'missing parameters' changeset

dwlehman:
__NOTE__: This will need to be rebased once the `safe_mode` changes are merged.

`Fix key for partition pool...`
This leads to a failure (crash?) any time a pool of type 'partition' is present.

`Don't unmount anything if only...`
This one is pretty self-explanatory.

`Add proper checking of specified disk...`
We already validate the supplied value of `disks` for all pools, but not for disk volumes.

`Fix passing of fs create options...`
We were incorrectly passing user-specified mkfs arguments to blivet, causing them to be ignored.

`Fail on request to resize unresizable...`
We should not silently fail if the user requests a resize of a non-resizable volume.

`Remove partitions etc as needed...`
This is _sort of_ a special-case for disk volumes, but it could be needed for other types as well.

`Fail w/ error if mount specified...`
Error out early if the user specifies mounting of a non-mountable formatting type.

`Add a default empty volume list to...`
This solves current crashes when a pool is defined with no `volumes` list.

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -167,7 +167,7 @@ class BlivetVolume(object):
         fmt = get_format(self._volume['fs_type'],
                          mountpoint=self._volume.get('mount_point'),
                          label=self._volume['fs_label'],
-                         options=self._volume['fs_create_options'])
+                         create_options=self._volume['fs_create_options'])
         if not fmt.supported or not fmt.formattable:
             raise BlivetAnsibleError("required tools for file system '%s' are missing" % self._volume['fs_type'])
 
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -251,6 +251,9 @@ class BlivetVolume(object):
         if self._device.exists:
             self._reformat()
 
+        if self.ultimately_present and self._volume['mount_point'] and not self._device.format.mountable:
+            raise BlivetAnsibleError("volume '%s' has a mount point but no mountable file system" % self._volume['name'])
+
         # schedule resize if appropriate
         if self._device.exists and self._volume['size']:
             self._resize()
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -209,6 +209,8 @@ class BlivetVolume(object):
                 raise BlivetAnsibleError("volume '%s' cannot be resized from %s to %s: %s" % (self._device.name,
                                                                                               self._device.size,
                                                                                               size, str(e)))
+        elif size and self._device.size != size and not self._device.resizable:
+            raise BlivetAnsibleError("volume '%s' cannot be resized from %s to %s" % (self._device.name, self._device.size, size))
 
     def _reformat(self):
         """ Schedule actions as needed to ensure the volume is formatted as specified. """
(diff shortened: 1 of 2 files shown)


library/blivet.py
```diff
@@ -258,6 +258,19 @@ def _get_device_id(self):
     def _type_check(self):
         return self._device.is_disk
 
+    def _look_up_device(self):
+        super(BlivetDiskVolume, self)._look_up_device()
+        if not self._get_device_id():
+            # FAIL: no disks specified for volume
```
tyll: The comment can be removed, it contains the same text as the exception in the next line

library/blivet.py
```diff
@@ -258,6 +258,19 @@ def _get_device_id(self):
...
+            raise BlivetAnsibleError("no disks specified for volume '%s'" % self._volume['name'])  # sure about this one?
+        elif not isinstance(self._volume['disks'], list):
+            raise BlivetAnsibleError("volume disks must be specified as a list")
+
+        if self._device is None:
+            # FAIL: failed to find the disk
```
tyll: The comment can be removed, it contains the same text as the exception in the next line

library/blivet.py
```diff
@@ -258,6 +258,19 @@ def _get_device_id(self):
     def _type_check(self):
         return self._device.is_disk
 
+    def _look_up_device(self):
+        super(BlivetDiskVolume, self)._look_up_device()
+        if not self._get_device_id():
+            # FAIL: no disks specified for volume
+            raise BlivetAnsibleError("no disks specified for volume '%s'" % self._volume['name'])  # sure about this one?
```
tyll: The comment at the end of this line is irritating. Is there an issue that should be tracked or can it be removed?

pcahyna:
@dwlehman this is ready to be merged, right?

pcahyna:
Do you please have tests for those problems?

pcahyna:
I suppose the test disabled in https://github.com/linux-system-roles/storage/pull/53/commits/136e41d134e31e6544b60078b71c45f7a596669c can be one of them

pcahyna:
maybe this : https://github.com/linux-system-roles/storage/pull/43#discussion_r334576458 is another instance?

library/blivet.py
```diff
@@ -209,6 +209,8 @@ def _resize(self):
                 raise BlivetAnsibleError("volume '%s' cannot be resized from %s to %s: %s" % (self._device.name,
                                                                                               self._device.size,
                                                                                               size, str(e)))
+        elif size and self._device.size != size and not self._device.resizable:
+            raise BlivetAnsibleError("volume '%s' cannot be resized from %s to %s" % (self._device.name, self._device.size, size))
```
pcahyna: Is it intended that the role fails in this case already in the "get required packages" task, before the actual resizing attempt?

pcahyna:
merging - but tests for bugs that it fixes are missing and will need to be added separately.

pcahyna:
I reverted master, this change breaks LUKS.

pcahyna:
Note: the rebased version is missing some commits, but that's because they are already in master. `Add proper checking of specified disk` (06b60aab4c1b834c7f636e6708e1c2c81c49e67e) is part of ab7389d8d31ec699b0cfcd43b3d98b47bc8b2052 (Add flag to prevent implicit removal of existing device/fs.)

PR #68, opened 2019-11-19
Manually manage swap entries in /etc/fstab.

dwlehman:
The 'mount' module removes pre-existing entries that match the
mount point of a new entry, which is problematic for swap since
they all have the same mount point: 'swap'.


dwlehman:
Closing in favor of #90

PR #70, opened 2020-01-14
Common fixes tests

pcahyna:
follows PR #64


pcahyna:
See #123 for actual tests for #64

PR #72, opened 2020-01-14
Make remaining class new-style (derive from object)

pcahyna:
For consistency - other classes needed this, otherwise super() failed with Python 2 (where classic classes are the default)


pcahyna:
was already committed in https://github.com/linux-system-roles/storage/commit/8697620df5373b7c882cb5bef6baeb80bb611d02 but partly lost in rebases

PR #74, opened 2020-03-25
sync with latest template; use tox

richm:
sync with latest template changes
use tox as primary test runner/driver
Note that this disables black and flake8 - should have
separate PRs to address that


richm:
nope, needs more template work

PR #76, opened 2020-03-29
use tox; use latest template - yamllint, etc.

richm:
use tox to run tests
use latest template - yamllint, and other changes


richm:
deleted in favor of https://github.com/linux-system-roles/storage/pull/78 which has the latest changes from network role feedback

PR #85, opened 2020-04-23
MDRAID support

japokorn:
- added mdraid support for volumes and pools
- added mdraid tests
- known issue: removing RAID volume created w/o using partition fails

### DIFF
--- a/tests/tests_create_raid_pool_then_remove.yml
+++ b/tests/tests_create_raid_pool_then_remove.yml
@@ -1,0 +1,89 @@
+---
+- hosts: all
+  become: true
+  vars:
+    storage_safe_mode: false
+    storage_use_partitions: true
+    mount_location1: '/opt/test1'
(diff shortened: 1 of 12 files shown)


tests/tests_create_raid_pool_then_remove.yml
```diff
@@ -0,0 +1,89 @@
...
+    volume2_size: '3g'
+
+  tasks:
+    - include_role:
+        name: storage
+
+    - include_tasks: get_unused_disk.yml
+      vars:
+        max_return: 2
+
```
yizhanglinux, line 21: do we need set min_size here, 
if only one disk retrurned, then the min_size should be 8+ g,

japokorn:
Work on this task has been split into two separate areas: MDRAID pools and MDRAID volumes. 
This PR is now obsolete, so I am closing it.
Updated PR related to pools is #91, the one for volumes hasn't been created yet.

PR #86, opened 2020-04-28
Streamline dependency gathering

dwlehman:
This is an untested concept of how we can substantially reduce the amount of code we run when gathering dependencies.

PR #87, opened 2020-04-29
Check for duplicate pool/volume names.

### LINKED ISSUE #83: No handle checking for the duplicate volume name
When I create the duplicated name lv, it will be failed and have the warning ouput.
When I test this scenario with playbook, I didn't find such warning information.

[root@storageqe-62 storage]# lvcreate  -L 3G -n test2 foo 
WARNING: xfs signature detected on /dev/foo/test2 at offset 0. Wipe it? [y/n]: y
  Wiping xfs signature on /dev/foo/test2.
  Logical volume "test2" created.
[root@storageqe-62 storage]# lvcreate  -L 3G -n test2 foo 
  Logical Volume "test2" already exists in volume group "foo"
[root@storageqe-62 storage]# echo $?
5

test case:
[root@storageqe-62 storage]# cat tests/test1.yml 
---
- hosts: all
  become: true
  vars:
    volume_group_size: '10g'
    volume_size: '3g'

  tasks:
    - include_role:
        name: storage

    - include_tasks: get_unused_disk.yml
      vars:
        min_size: "{{ volume_group_size }}"
        max_return: 1

    - name: Create three LVM logical volumes under one volume group
      include_role:
        name: storage
      vars:
        storage_pools:
            - name: foo
              disks: "{{ unused_disks }}"
              volumes:
                - name: test1
                  size: "{{ volume_size }}"
                  mount_point: '/opt/test1'
                - name: test2
                  size: "{{ volume_size }}"
                  mount_point: '/opt/test2'
                - name: test2
                  size: "{{ volume_size }}"
                  mount_point: '/opt/test3'

Executing log:
TASK [Print out pool information] ********************************************************************************************************************************************************************************
task path: /root/storage/tests/verify-role-results.yml:1
ok: [localhost] => {
    "_storage_pools_list": [
        {
            "disks": [
                "sdb"
            ],
            "name": "foo",
            "state": "present",
            "type": "lvm",
            "volumes": [
                {
                    "_device": "/dev/mapper/foo-test1",
                    "_mount_id": "/dev/mapper/foo-test1",
                    "fs_create_options": "",
                    "fs_label": "",
                    "fs_overwrite_existing": true,
                    "fs_type": "xfs",
                    "mount_check": 0,
                    "mount_device_identifier": "uuid",
                    "mount_options": "defaults",
                    "mount_passno": 0,
                    "mount_point": "/opt/test1",
                    "name": "test1",
                    "pool": "foo",
                    "size": "3g",
                    "state": "present",
                    "type": "lvm"
                },
                {
                    "_device": "/dev/mapper/foo-test2",
                    "_mount_id": "/dev/mapper/foo-test2",
                    "fs_create_options": "",
                    "fs_label": "",
                    "fs_overwrite_existing": true,
                    "fs_type": "xfs",
                    "mount_check": 0,
                    "mount_device_identifier": "uuid",
                    "mount_options": "defaults",
                    "mount_passno": 0,
                    "mount_point": "/opt/test2",
                    "name": "test2",
                    "pool": "foo",
                    "size": "3g",
                    "state": "present",
                    "type": "lvm"
                },
                {
                    "_device": "/dev/mapper/foo-test2",
                    "_mount_id": "/dev/mapper/foo-test2",
                    "fs_create_options": "",
                    "fs_label": "",
                    "fs_overwrite_existing": true,
                    "fs_type": "xfs",
                    "mount_check": 0,
                    "mount_device_identifier": "uuid",
                    "mount_options": "defaults",
                    "mount_passno": 0,
                    "mount_point": "/opt/test3",
                    "name": "test2",
                    "pool": "foo",
                    "size": "3g",
                    "state": "present",
                    "type": "lvm"
                }
            ]
        }
    ]
}

dwlehman:
Fixes #83.

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -125,6 +125,22 @@ disklabel_type = None  # user-specified disklabel type
 safe_mode = None       # do not remove any existing devices or formatting
 
 
+def find_duplicates(dicts):
+    """ Return a list of names that appear more than once in a list of dicts.
+
+        Items can be a list of any dicts with a 'name' key; that's all we're
+        looking at. """
+    names = list()
+    duplicates = list()
+    for item in dicts:
+        if item['name'] in names:
+            duplicates.append(item['name'])
+        else:
+            names.append(item['name'])
+
+    return duplicates
+
+
 class BlivetAnsibleError(Exception):
     pass
 
@@ -780,12 +796,25 @@ def run_module():
                     fs_type=action.format.type if action.is_format else None,
                     device=action.device.path)
 
+    duplicates = find_duplicate_names(module.params['pools'])
+    if duplicates:
+        module.fail_json(msg="multiple pools with the same name: {0}".format(",".join(duplicates)),
+                         **result)
     for pool in module.params['pools']:
+        duplicates = find_duplicate_names(pool['volumes'])
+        if duplicates:
+            module.fail_json(msg="multiple volumes in pool '{0}' with the "
+                                 "same name: {1}".format(pool['name'], ",".join(duplicates)),
+                             **result)
         try:
             manage_pool(b, pool)
         except BlivetAnsibleError as e:
             module.fail_json(msg=str(e), **result)
 
+    duplicates = find_duplicate_names(module.params['volumes'])
+    if duplicates:
+        module.fail_json(msg="multiple volumes with the same name: {0}".format(",".join(duplicates)),
+                         **result)
     for volume in module.params['volumes']:
         try:
             manage_volume(b, volume)
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -133,7 +133,7 @@ def find_duplicate_names(dicts):
     names = list()
     duplicates = list()
     for item in dicts:
-        if item['name'] in names:
+        if item['name'] in names and item['name'] not in duplicates:
             duplicates.append(item['name'])
         else:
             names.append(item['name'])
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -125,7 +125,7 @@ disklabel_type = None  # user-specified disklabel type
 safe_mode = None       # do not remove any existing devices or formatting
 
 
-def find_duplicates(dicts):
+def find_duplicate_names(dicts):
     """ Return a list of names that appear more than once in a list of dicts.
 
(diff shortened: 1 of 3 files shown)


dwlehman:
Funny enough, the tests here are failing because of issues solved by #86 -- namely, the 'get packages' run fails without setting `blivet_output`, causing the test validation to fail.

library/blivet.py
```diff
@@ -124,6 +124,22 @@
 packages_only = None   # only set things up enough to get a list of required packages
 
 
+def find_duplicate_names(dicts):
+    """ Return a list of names that appear more than once in a list of dicts.
+
+        Items can be a list of any dicts with a 'name' key; that's all we're
+        looking at. """
+    names = list()
```
richm, line 132: would be faster to use `names = set()` since you only use `names` to see if this is a duplicate or not - then `names.add(item['name'])` instead of `names.append(item['name'])`

dwlehman: That's true. I don't think it will matter in practice since it's unlikely users will have more than 10s of pools or volumes at any given level. It would also make sense for `duplicates` to be a set, but I think the randomized ordering is probably worse than the sub-optimal efficiency (and sorting or casting to a list would presumable wash the performance gains of using a set in the first place. So, while I don't mind making the change, I am doubtful it will have any real-world effect.

richm: ok - then leave it as is - I thought `duplicates` was a list so that you could show the user how many duplicates there were e.g. the output would be `volA volA volA volB volB` or such, which might be more useful than just `volA volB`, and doing something like `volA:3 volB:2` e.g. show the count of each one is also probably a bit much for 10s of items.

dwlehman: Hah! Now you caught a bug. My intention was to only have each name appear once in `duplicates`, but I didn't write it that way. I'm going to fix that.

library/blivet.py
```diff
@@ -132,7 +132,7 @@ def find_duplicate_names(dicts):
     names = list()
     duplicates = list()
     for item in dicts:
-        if item['name'] in names:
+        if item['name'] in names and item['name'] not in duplicates:
```
richm, line 135: ship it

PR #88, opened 2020-05-06
Avoid using ansible_failed_task in rescue blocks

pcahyna:
`ansible_failed_task` can contain unexpanded Jinja2 templates with variable references that were valid when the task failed, but not valid anymore when using the variable in the `rescue` block. In this case the problem is with task names (`.name`) that contain references to `storage_provider`. Unfortunately this completely prevents using `ansible_failed_task` even if not using the problematic attribute. Other attribute could have been used if this were not the case, like `.tags`, but until https://github.com/ansible/ansible/issues/49942 is fixed, it is better to avoid using `ansible_failed_task` and use the only other special variable describing the failed task: `ansible_failed_result`.

Eliminate the previous workaround to set storage_provider on the tasks that expand `ansible_failed_task` and use `ansible_failed_result` with a special value of `msg` instead. This has a disadvantage of having to use a special `msg` with the `fail` module, while usually it is just a free-form message. The `fail` module does not set another attribute of the result that could be used though, and it is cleaner than the previous workaround.

### DIFF
--- a/tests/tests_disk_errors.yml
+++ b/tests/tests_disk_errors.yml
@@ -26,19 +26,16 @@
                 disks: ['/dev/surelyidonotexist']
                 mount_point: "{{ mount_location }}"
 
-        - name: UNREACH
+        - name: unreachable task
           fail:
-            msg: "this should be unreachable"
+            msg: UNREACH
 
       rescue:
         - name: Check that we failed in the role
           assert:
             that:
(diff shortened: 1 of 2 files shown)


yizhanglinux:
lgtm

pcahyna:
Note to myself: this trick should be added to https://github.com/oasis-roles/meta_standards/pull/19

Issue #66, opened 2019-11-11
fstype: swap not working 


parul1593:
Is swap supported as fs_type? Per documentation I see it is supported but when I run playbook, it says unknown fs_type.

I am using below playbook to create swap volume:
---
- hosts: localhost
  become: true
#  vars:
#    storage_use_partitions: true

  roles:
    - name: linux-system-roles.storage
      storage_pools:
        - name: app_data_01
          disks: ['/dev/sdb']
          #state: "absent"
          volumes:
            - name: data1
              fs_label: data1
              fs_type: ext4
              size: 10g
              mount_point: '/fs/data1'
        - name: data_02
          disks: ['/dev/sdc','/dev/sdd']
          #state: "absent"
          volumes:
            - name: data2
              size: 8g
              fs_type: swap
              mount_point: /fs/hdata2
            - name: data3
              mount_options: 'ro,noatime'
              fs_type: xfs
              size: 2g
              mount_point: /fs/data3

This is the error which I get
"msg": "Error mounting /fs/hdata2: mount: unknown filesystem type 'swap'\n"

dwlehman:
Swap cannot have a mount point. It does not actually get mounted; it gets activated instead.

parul1593:
If I dont give a mountpoint, then it goes through fine. But there is no entry in /etc/fstab and I dont see any swap getting added. I ran "swapon -a" also.

Using this playbook:
`---
- hosts: localhost
  become: true
#  vars:
#    storage_use_partitions: true

  roles:
    - name: linux-system-roles.storage
      storage_pools:
        - name: app_data_01
          disks: ['/dev/sdb']
          #state: "absent"
          volumes:
            - name: data1
              fs_label: data1
              fs_type: ext4
              size: 10g
              mount_point: '/fs/data1'
        - name: app_data_02
          disks: ['/dev/sdc','/dev/sdd']
          #state: "absent"
          volumes:
            - name: data2
              size: 4g
              fs_type: ext4
              mount_point: /fs/data2
              #mount_options: 'noatime'
              #type: lvm
              #fs_overwrite_existing: false
            - name: data3
              #mount_options: 'ro,noatime'
              fs_type: swap
              size: 15g
`

output of lsblk:
NAME                 MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
fd0                    2:0    1    4K  0 disk
sda                    8:0    0   64G  0 disk
|-sda1                 8:1    0  512M  0 part /boot/efi
[...]

PR #90, opened 2020-05-14
Correctly Manage Swap Volumes

dwlehman:
This should address #66.

It changes less than #68 did, since some of those changes were based on incorrect assertions, namely that the mount module doesn't support removing specific swap entries from /etc/fstab.

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -733,6 +733,26 @@ def get_required_packages(b, pools, volumes):
     return sorted(list(set(packages)))
 
 
+def update_fstab_identifiers(b, pools, volumes):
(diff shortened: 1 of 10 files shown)


dwlehman:
"Fix conditions for removing partition pool in _create" is technically unrelated, but was causing CI failures. If desired, I can split it out into a separate pull request, although I'm not clear on the benefits of doing so.

dwlehman:
The missing `libblockdev-swap` package only appeared to be an issue on RHEL-7, which I do not understand (why not also RHEL-8 and Fedora?).

dwlehman:
[citest bad]

PR #91, opened 2020-05-18
MDRAID support for pools

japokorn:
added mdraid support for pools
added mdraid tests


library/blivet.py
```diff
@@ -482,17 +511,17 @@ def _look_up_device(self):
...
-        for disk in self._disks:
+        for disk in disks:
             if not disk.isleaf or disk.format.type is not None:
                 if safe_mode:
                     raise BlivetAnsibleError("cannot remove existing formatting and/or devices on disk '%s' (pool '%s') in safe mode" % (disk.name, self._pool['name']))
                 else:
                     self._blivet.devicetree.recursive_remove(disk)
 
-            if use_partitions:
+            if use_partitions and self._pool.get('raid_level') in [None, "null", ""]:
```
dwlehman, line 524: This is the logical place to call a function that creates the mdraid layer.

library/blivet.py
```diff
@@ -503,7 +532,7 @@ def _create_members(self):
             self._blivet.format_device(member, self._get_format())
             members.append(member)
 
-        if use_partitions:
+        if use_partitions and self._pool.get('raid_level') in [None, "null", ""]:
```
dwlehman, line 535: It looks like it might be worthwhile to add a property called `is_raid` or similar to streamline this check since it's repeated several times.

library/blivet.py
```diff
@@ -559,6 +590,40 @@ def _create(self):
             label = get_format("disklabel", device=self._device.path, label_type=disklabel_type)
             self._blivet.format_device(self._device, label)
 
+class BlivetMDRaidPool(BlivetPool):
+    def create(self):
+        raid_name = "mdraid-%s" % self._pool['name']
+        member_names = self._pool['disks']
```
dwlehman, line 596: There should already be a method to get the blivet disk instances corresponding to the identifiers in pool['disks'], which can contain any of several identifiers, including UUID=<uuid>, /dev/disk/by-path/<link>, etc. I believe they are accessible as `self._disks`.

library/blivet.py
```diff
@@ -559,6 +590,40 @@ def _create(self):
...
+
+        # valid member count
+        if raid_level_obj.min_members < len(member_names):
+            raise BlivetAnsibleError("at least %s members are required for RAID level '%s'" % (raid_level_obj.min_members, raid_level))
+
+        if safe_mode:
+            raise BlivetAnsibleError("cannot create new RAID '%s' in safe mode" % safe_mode)
+        raid_members = create_raid_members(self._blivet, member_names)
+
+        raid_array = MDRaidArrayDevice(name=raid_name, level=raid_level, member_devices=len(raid_members),
```
dwlehman, line 615: It would be a bit simpler if you followed the convention of the other pool/volume classes and used `self._blivet.new_mdarray` instead of importing and instantiating the class directly.

tests/test-verify-volume-mount.yml
```diff
@@ -1,9 +1,27 @@
 ---
+
+#
+# MDRaid devices paths are returned as a symlinks. But we need their targets.
+#
```
dwlehman, line 5: Since this block appears in several places it should instead go into tests/verify-role-results.yml in the block titled 'Collect some information about the current state of the system."

You can also get the real/system device node by doing this:

```
- command: realpath "{{ storage_test_volume._device }}"
  register: storage_test_sys_node
```

(I did exactly this in the test result validation for swap in #90.)

library/blivet.py
```diff
@@ -559,6 +590,40 @@ def _create(self):
             label = get_format("disklabel", device=self._device.path, label_type=disklabel_type)
             self._blivet.format_device(self._device, label)
 
+class BlivetMDRaidPool(BlivetPool):
```
dwlehman, line 593: I don't think this should be a subclass of `BlivetPool`, but I do think it might make sense to have some shared code for creating the arrays. The same will be true for encryption.

library/blivet.py
japokorn: Fixed

library/blivet.py
japokorn: BlivetMDRaidPool is no longer a subclass of BlivetPool.

library/blivet.py
japokorn: Fixed: `self._disks` is now obtained as constructor parameter since `BlivetMDRaidPool` no longer inherits from `BlivetPool`

library/blivet.py
japokorn: Fixed: Now using `self._blivet.new_mdarray`. Removed `MDRaidArrayDevice` import.

tests/test-verify-volume-mount.yml
japokorn: Partially fixed: Streamlined the code to use `realpath` instead. This part of code may be affected by upcoming mdraid volume PR which sometimes requires keeping symlink paths. I would rather do proposed change in that PR.

library/blivet.py
```diff
@@ -559,6 +591,49 @@ def _create(self):
...
+                                               parents=raid_members)
+
+
+        if not safe_mode:
+            self._blivet.create_device(raid_array)
+
+            try:
+                do_partitioning(self._blivet)
+            except Exception:
+                raise BlivetAnsibleError("failed to allocation partitions for pool '%s'" % self._pool['name'])
```
dwlehman, line 634: I don't think this class serves any purpose as it is now. The only thing I think it adds is the `new_mdarray` call, which could easily go into `BlivetPool._create_members`.

library/blivet.py
```diff
@@ -559,6 +591,49 @@ def _create(self):
...
+        member_names = [x.name for x in self._disks]
+        raid_level = self._pool['raid_level']
+
+        # check validity of argument values
+
+        # valid RAID level (e.g. "raid0", "raid1", "linear"...)
+        try:
+            raid_level_obj = get_raid_level(raid_level)
+        except RaidError:
+            raise BlivetAnsibleError("invalid RAID level value '%s'" % raid_level)
```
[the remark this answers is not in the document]
japokorn: Fixed: Removed the check and added `ValueError` exception catching to `new_mdarray` call to produce more descriptive message.

library/blivet.py
```diff
@@ -559,6 +591,49 @@ def _create(self):
...
+
+        # valid member count
+        if raid_level_obj.min_members < len(member_names):
+            raise BlivetAnsibleError("at least %s members are required for RAID level '%s'" % (raid_level_obj.min_members, raid_level))
```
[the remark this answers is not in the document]
japokorn: Fixed: Removed the check and added `ValueError` exception catching to `new_mdarray` call to produce more descriptive message.

library/blivet.py
```diff
@@ -500,16 +507,41 @@ def _create_members(self):
...
+
         if use_partitions:
             try:
                 do_partitioning(self._blivet)
             except Exception:
-                raise BlivetAnsibleError("failed to allocation partitions for pool '%s'" % self._pool['name'])
+                raise BlivetAnsibleError("failed to allocate partitions for pool '%s'" % self._pool['name'])
+
+        if self._is_raid:
+            self._blivet.format_device(raid_array, self._get_format())
```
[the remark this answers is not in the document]
japokorn: Fixed: Formatting added to `new_mdarray` call.

library/blivet.py
```diff
@@ -756,6 +790,7 @@ def run_module():
 
     module = AnsibleModule(argument_spec=module_args,
                            supports_check_mode=True)
+
```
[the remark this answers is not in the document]
japokorn: Fixed: Empty line removed.

library/blivet.py
```diff
@@ -540,6 +572,7 @@ def manage(self):
         self._manage_volumes()
 
 
+
```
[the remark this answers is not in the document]
japokorn: Fixed: Empty line removed.

README.md
```diff
@@ -30,6 +30,10 @@ Supported identifiers include: device node (like `/dev/sda` or `/dev/mapper/mpat
 device node basename (like `sda` or `mpathb`), /dev/disk/ symlink
 (like `/dev/disk/by-id/wwn-0x5000c5005bc37f3f`).
 
+##### `raid_level`
+When used with `type: lvm` it manages a volume group with a mdraid array of given level
+on it. Accepted values are: `linear`, `striped`, `raid0`, `raid1`, `raid4`, `raid5`, `raid6`, `raid10`
+
```
pcahyna, line 36: I am missing here a description on how to specify member drives of a RAID.
Normally, the `disks` are a list of disks to be used as PVs of the volume group. But in the case of RAID, they change their meaning: there will be a single PV and the `disks` determine the components of the RAID. This is not completely intuitive and warrants a better description.
Also, `raid10` is two-level - how do you specify which disks belong together in mirrors? https://en.wikipedia.org/wiki/Nested_RAID_levels#RAID_10_.28RAID_1.2B0.29

japokorn: Fixed: Added explicit note about `disks` behavior when `raid_level` option is used.

RAID handling is passed to blivet. At this point there is no additional logic regarding `raid10`.

pcahyna: What does blivet do for raid10, then?

dwlehman: The layout (which could perhaps use a brief explanation in README.md), is the usual arrangement: a single array is created across the specified disks and used as the lone member device for the pool.

The details of a RAID10 layout are handled by `mdadm(8)`. There is currently no way to override mdadm's default for that.

PR #95, opened 2020-05-24
Avoid using ansible_failed_task in rescue blocks second round

yizhanglinux:


yizhanglinux:
[citest]

PR #100, opened 2020-06-05
Encrypted Volumes

dwlehman:
This adds support for LUKS volumes. ~~The testing is very minimal. I'm curious what it will take to get it to pass CI.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,27 @@ The `mount_point` specifies the directory on which the file system will be mount
 ##### `mount_options`
 The `mount_options` specifies custom mount options as a string, e.g.: 'ro'.
 
+##### `encryption`
+This specifies whether or not the volume will be encrypted using LUKS.
+__WARNING__: Toggling encryption for a volume is a destructive operation, meaning
+             all data on that volume will be removed as part of the process of
+             adding/removing the encryption layer.
+
+##### `encryption_passphrase`
+This string specifies a passphrase used to unlock/open the LUKS volume.
+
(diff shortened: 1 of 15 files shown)


dwlehman:
This is built on top of my swap-minimal branch (from #90). Only the last five commits are specific to this.

richm:
@sergio-correia FYI

dwlehman:
Weird that the luks tests should be what fails since that's what I ran locally on F32 prior to opening/updating this pull request.

dwlehman:
Oh, good grief. It only passes on the OS I developed and initially tested on? It'll have to wait til next week.

dwlehman:
What's missing at this point is unknown, aside from better testing:

1. verify that the LUKS layer exists as appropriate and has the expected name
2. verify the LUKS parameters insofar as that is possible

README.md
```diff
@@ -73,6 +73,27 @@ The `mount_point` specifies the directory on which the file system will be mount
...
+This string specifies a passphrase used to unlock/open the LUKS volume.
+
+##### `encryption_key_file`
+This string specifies the full path to the key file used to unlock the LUKS volume.
+
+##### `encryption_cipher`
+This string specifies a non-default cipher to be used by LUKS.
+
+##### `encryption_key_size`
+This integer specifies the LUKS key size (in bytes).
```
vojtechtrefny, line 92: The key size should be in bits, e.g. 512

dwlehman: Doh, good catch -- thanks!

PR #102, opened 2020-06-08
Encrypted Pools

dwlehman:
This seems fairly complete ~~, but the tests are woefully inadequate~~ . This is on top of #90 and #100. The relevant work begins with a3293e9.

Edit: tests are reasonable now, although they don't check cipher, key size, or luks version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,27 @@ Accepted values are: `linear`, `striped`, `raid0`, `raid1`, `raid4`, `raid5`, `r
 This is a list of volumes that belong to the current pool. It follows the
 same pattern as the `storage_volumes` variable, explained below.
 
+##### `encryption`
+This specifies whether or not the pool will be encrypted using LUKS.
+__WARNING__: Toggling encryption for a pool is a destructive operation, meaning
+             the pool itself will be removed as part of the process of
+             adding/removing the encryption layer.
+
+##### `encryption_passphrase`
+This string specifies a passphrase used to unlock/open the LUKS volume(s).
+
+##### `encryption_key_file`
+This string specifies the full path to the key file used to unlock the LUKS volume(s).
+
+##### `encryption_cipher`
+This string specifies a non-default cipher to be used by LUKS.
+
+##### `encryption_key_size`
+This integer specifies the LUKS key size (in bytes).
+
+##### `encryption_luks_version`
+This integer specifies the LUKS version to use.
+
 
 #### `storage_volumes`
 The `storage_volumes` variable is a list of volumes to manage. Each volume has the following
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -538,6 +538,8 @@ class BlivetPool(BlivetBase):
 
             leaves = [a for a in ancestors if a.isleaf]
 
+        self._device = None
+
     def _type_check(self):  # pylint: disable=no-self-use
         return True
 
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -148,14 +148,26 @@ class BlivetAnsibleError(Exception):
     pass
 
 
-class BlivetVolume(object):
+class BlivetBase(object):
     blivet_device_class = None
+    _type = None
+
+    def __init__(self, blivet_obj, spec_dict):
+        self._blivet = blivet_obj
+        self._spec_dict = spec_dict
+        self._device = None
+
+
+class BlivetVolume(BlivetBase):
+    _type = "volume"
 
     def __init__(self, blivet_obj, volume, bpool=None):
-        self._blivet = blivet_obj
-        self._volume = volume
+        super(BlivetVolume, self).__init__(blivet_obj, volume)
         self._blivet_pool = bpool
-        self._device = None
+
+    @property
+    def _volume(self):
+        return self._spec_dict
 
     @property
     def required_packages(self):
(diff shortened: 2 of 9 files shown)


dwlehman:
The obvious things missing from testing:

1. verification of absence/presence of LUKS layer as appropriate
2. correct application of various parameters insofar as that is possible

library/blivet.py
```diff
@@ -145,15 +148,72 @@ class BlivetAnsibleError(Exception):
     pass
 
 
-class BlivetVolume(object):
+class BlivetBase(object):
```
richm, line 151: do you need unit tests for these changes?  if so, they can be added later after the 15th

tests/tests_swap.yml
```diff
@@ -0,0 +1,91 @@
...
+        name: storage
+      vars:
+        storage_volumes:
+          - name: test1
+            type: disk
+            mount_point: "{{ mount_location }}"
+            fs_type: ext3
+            disks: "{{ unused_disks }}"
+
+    - include_tasks: verify-role-results.yml
```
richm, line 54: Does `verify-role-results.yml` check for idempotence?  Does it do something like `assert: that: some_register_var is not changed`?

dwlehman: No, the only tests we have for idempotence are explicitly for that purpose -- basically running the role a second time and checking that things are the same afterwards. I don't think I did any of those for this changeset, though. Maybe I'll add one in the morning.

pcahyna:
@sergio-correia the LUKS parameters are needlessly different from the ones used in your role. https://github.com/linux-system-roles/nbde_client/pull/3/files#diff-7eeda618087b49ae876084ab6c73fdbbR9 uses `pass` and `keyfile`. Here one uses `encryption_passphrase` and `encryption_key_file`. Those names should be consistent. (I don't care which one is chosen.)

README.md
```diff
@@ -39,6 +39,27 @@ Accepted values are: `linear`, `striped`, `raid0`, `raid1`, `raid4`, `raid5`, `r
...
+This specifies whether or not the pool will be encrypted using LUKS.
+__WARNING__: Toggling encryption for a pool is a destructive operation, meaning
+             the pool itself will be removed as part of the process of
+             adding/removing the encryption layer.
+
+##### `encryption_passphrase`
+This string specifies a passphrase used to unlock/open the LUKS volume(s).
+
+##### `encryption_key_file`
+This string specifies the full path to the key file used to unlock the LUKS volume(s).
```
pcahyna, line 52: path on the control node or on the managed host?

dwlehman: Everything is on the managed host AFAIK. Any operations to get the key file to the managed nodes is on the user. Maybe I should note that.

sergio-correia: > @sergio-correia the LUKS parameters are needlessly different from the ones used in your role. https://github.com/linux-system-roles/nbde_client/pull/3/files#diff-7eeda618087b49ae876084ab6c73fdbbR9 uses `pass` and `keyfile`. Here one uses `encryption_passphrase` and `encryption_key_file`. Those names should be consistent. (I don't care which one is chosen.)

I will adopt `encryption_passphrase` and `encryption_key_file`. These seem more explicit.

pcahyna: I think you an use just `passphrase` and `keyfile`. In your role it is obvious that it is encryption related (it is an encryption role). I was concerned about the gratuitous `pass` vs `passphrase` difference or `keyfile` vs. `key_file`.

dwlehman:
@pcahyna @sergio-correia we use the prefixes for a sort of namespacing, so all the encryption-specific parameters have the 'encryption_' prefix. Same for raid, fs, mount, etc. Not sure if there's a benefit for the clevis/tang roles to do something along the same lines.

sergio-correia:
> @pcahyna @sergio-correia we use the prefixes for a sort of namespacing, so all the encryption-specific parameters have the 'encryption_' prefix. Same for raid, fs, mount, etc. Not sure if there's a benefit for the clevis/tang roles to do something along the same lines.

Right, I wont use the prefix. @pcahyna explained he was more concerned about pass/passphrase (we prefer passphrase, so I will update there) and keyfile/key_file

README.md
pcahyna: @dwlehman would it make sense to have `key_file` on the control node? `nbde_client` uses a keyfile on the control node: https://github.com/linux-system-roles/nbde_client/pull/3#discussion_r439006359 . Are the use cases different enough to justify this difference in semantics?

dwlehman: I agree that it is more convenient for users and the consistency between roles is good.

Where on the managed node should we place the key files?

pcahyna: > Where on the managed node should we place the key files?

@sergio-correia what do you do in nbde_client?

sergio-correia: I define an internal variable `__nbde_client_managed_dir` and have a task to copy the key files there and do the cleanup later. I pass this variable also to the python module, so it knows where to find the key files. It's currently defined as `/var/run/linux-system-roles.nbde_client`

dwlehman: For our purposes I think we need to keep the key file around for reboots, while I (think that I) see that you are removing them in the nbde_client role after setting up nbde. For us to specify that they be on the controller necessitates developing a scheme for storing them on the managed nodes.

sergio-correia: Yeah, for us it's simpler that we need it to do some operations, then we don't need it anymore and can remove it.

PR #107, opened 2020-06-14
Update supported Fedora versions

pcahyna:
Will also serve to rerun CI after the recent merges.


pcahyna:
@japokorn , your tests_create_raid_volume_then_remove.yml now reliably fails, please have a look. (tests_raid_volume_options.yml fails in a similar way).

pcahyna:
Merging of #102 broke it.

pcahyna:
The problem was fixed by #109 .

PR #114, opened 2020-06-17
raid fix - removed conflicting defaults

japokorn:
- removed default values for chunk size and metadata version

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -96,7 +96,6 @@ try:
     from blivet3 import Blivet
     from blivet3.callbacks import callbacks
     from blivet3 import devices
-    from blivet3.devicelibs.mdraid import MD_CHUNK_SIZE
     from blivet3.flags import flags as blivet_flags
     from blivet3.formats import get_format
     from blivet3.partitioning import do_partitioning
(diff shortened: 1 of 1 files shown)


library/blivet.py
```diff
@@ -526,8 +524,8 @@ def _create(self):
                                                                    self._volume.get("raid_device_count"),
                                                                    self._volume.get("raid_spare_count"))
 
-        chunk_size = Size(self._volume.get("raid_chunk_size", MD_CHUNK_SIZE))
-        metadata_version = self._volume.get("raid_metadata_version", "default")
+        chunk_size = Size(self._volume.get("raid_chunk_size"))
```
dwlehman, line 527: I don't think this will go well. Shouldn't it be `Size(...)` only if it has a size-like value, otherwise `None` to let blivet pick the default?

japokorn: True.
Fixed: None value will now prevent casting to Size. Added handling None value later in code.

pcahyna:
Does this fix a user-visible bug?

japokorn:
@pcahyna: The default values in `defaults/main.yml` were overwritten by defaults of the blivet. Modifying default values in the file would not work for raid. This fixes that issue.

pcahyna:
Thanks @japokorn . Is there a test for the bug?

japokorn:
No. This is a small change.

pcahyna:
@japokorn the change may be small, but the bug not necessarily so. If `metadata_version` and `chunk_size` were ignored and the problem was not detected, it indicates a gap in test coverage.

japokorn:
@pcahyna It didn't show up on tests because under the standard circumstances are both default values the same.

japokorn:
This PR was obsoleted by #116

PR #115, opened 2020-06-22
Add tests for the optional encryption parameters

dwlehman:


tests/tests_luks.yml
```diff
@@ -15,6 +15,20 @@
...
 
+    - name: Create a key file
+      tempfile:
+        prefix: storage_test
+        suffix: lukskey
+      register: storage_test_key_file
+      changed_when: false
+
+    - name: Write the key into the key file
+      lineinfile:
```
richm, line 26: I would suggest using `copy` instead of `lineinfile` (the current best practice is to avoid `lineinfile` if possible - https://github.com/oasis-roles/meta_standards#ansible-best-practices)
```yaml
     - name: Write the key into the key file
       copy:
         dest: "{{ storage_test_key_file.path }}"
         content: veryveryveryveryverysecureindeed
       changed_when: false
```

tests/tests_luks.yml
```diff
@@ -285,3 +302,9 @@
             state: absent
 
     - include_tasks: verify-role-results.yml
+
+    - name: Remove the key file
```
richm, line 306: Suggest using `tempfile` in a `block` and doing the removal in the `always` section e.g. https://github.com/linux-system-roles/nbde_client/blob/master/tasks/main-clevis.yml#L7

tests/tests_luks.yml
dwlehman: Fixed in latest version in both `tests_luks.yml` and `tests_luks_pool.yml`.

tests/tests_luks.yml
dwlehman: Fixed in latest version in both `tests_luks.yml` and `tests_luks_pool.yml`.

tests/tests_luks_pool.yml
```diff
@@ -68,22 +54,44 @@
...
+      block:
+        - name: Create a key file
+          tempfile:
+            prefix: storage_test
+            suffix: lukskey
+          register: storage_test_key_file
+          changed_when: false
+
+        - name: Write the key into the key file
+          lineinfile:
```
richm, line 67: except for `lineinfile` lgtm

dwlehman: doh!

PR #116, opened 2020-06-24
Advanced options support for raid pools

japokorn:
- added support for additional raid pools options
- streamlined raid options handling for volumes
- added new test for raid pool options
- fixed typo in comment of existing raid test

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -96,7 +96,6 @@ try:
     from blivet3 import Blivet
     from blivet3.callbacks import callbacks
     from blivet3 import devices
-    from blivet3.devicelibs.mdraid import MD_CHUNK_SIZE
     from blivet3.flags import flags as blivet_flags
     from blivet3.formats import get_format
     from blivet3.partitioning import do_partitioning
@@ -109,7 +108,6 @@ except ImportError:
         from blivet import Blivet
         from blivet.callbacks import callbacks
         from blivet import devices
-        from blivet.devicelibs.mdraid import MD_CHUNK_SIZE
         from blivet.flags import flags as blivet_flags
         from blivet.formats import get_format
         from blivet.partitioning import do_partitioning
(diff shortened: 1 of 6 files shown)


pcahyna:
Is there any relation to #114 ?

japokorn:
@pcahyna This PR basically obsoletes #114 because it changes the code affected by it. I was planning to close #114 unless there is a reason not to.

library/blivet.py
```diff
@@ -146,6 +144,55 @@ def find_duplicate_names(dicts):
 
     return duplicates
 
+def new_mdarray(blivet, members, param_dict, fmt, raid_name=""):
```
dwlehman, line 147: This should go in `BlivetBase`, which I made for exactly this type of application.

tests/tests_raid_pool_options.yml
```diff
@@ -0,0 +1,72 @@
...
+                size: "{{ volume1_size }}"
+                mount_point: "{{ mount_location1 }}"
+              - name: lv2
+                size: "{{ volume2_size }}"
+                mount_point: "{{ mount_location2 }}"
+              - name: lv3
+                size: "{{ volume3_size }}"
+                mount_point: "{{ mount_location3 }}"
+
+    - include_tasks: verify-role-results.yml
```
dwlehman, line 46: I don't think there is anything to validate that the optional parameters were applied correctly.

library/blivet.py
japokorn: Fixed: Moved function to `BlivetBase` class, removed parameters obtainable from class.

tests/tests_raid_pool_options.yml
japokorn: Fixed: Added new file with raid pool options tests.

library/blivet.py
```diff
@@ -205,6 +203,56 @@ def _manage_one_encryption(self, device):
 
         return ret
 
+    def _new_mdarray(self, members, fmt, raid_name=""):
```
dwlehman, line 206: Both places you call this you are passing `self._get_format()` as `fmt`, so I think you can remove that parameter.

tests/test-verify-pool-md.yml
```diff
@@ -0,0 +1,50 @@
...
+  # Only when pool is raid
+- name: check RAID options
+  block:
+
+    - name: get information about RAID
+      command: "mdadm --detail /dev/md/{{ storage_test_pool.name }}-1"
+      register: storage_test_mdadm
+      changed_when: false
+      # Note: using "mdadm --detail {{ storage_test_pool.volumes[0]._device }}"
+      # causes mdadm segfault on RHEL even though it works on Fedora
```
dwlehman, line 12: What is the value of `storage_test_pool.volumes[0]._device` when it segfaults? Is that a valid path to the device?

japokorn: Yes. The whole command looked like this:
`mdadm --detail /dev/mapper/vg1-lv1`
It worked for Fedora but segfaulted on all RHELs.

dwlehman: Ah, because that is the wrong device. You want to pass it the PV device, not the LV device. You can see an example of listing the PVs in `tests/test-verify-pool-members.yml`.

library/blivet.py
japokorn: Fixed: Removed `fmt` parameter. Value is now obtained from `self`.

tests/test-verify-pool-md.yml
vojtechtrefny: Segfault is not a very good input validation :-) If you can reproduce this behaviour, please report it.

dwlehman: To be clear, this will work as-is for stacks created by the role, but it will not work for stacks created by the customer/user prior to running the role.

japokorn: @dwlehman: The fix is intended as a quick workaround. Since this call happens only within our test it should be safe.

dwlehman: Oh, that's right. This is only in the tests.

pcahyna:
@japokorn , Travis says : "The build failed. This is a change from the previous build, which passed." The failure is:
```
pylint run-test: commands[0] | bash /home/travis/build/linux-system-roles/storage/.travis/runpylint.sh --errors-only
+ python /home/travis/build/linux-system-roles/storage/.travis/custom_pylint.py --errors-only
/home/travis/build/linux-system-roles/storage/.travis/custom_pylint.py: files to be checked:
    /home/travis/build/linux-system-roles/storage/tests/unit/resolve_blockdev_test.py
    /home/travis/build/linux-system-roles/storage/tests/unit/gensym_test.py
    /home/travis/build/linux-system-roles/storage/tests/unit/test_unused_disk.py
    /home/travis/build/linux-system-roles/storage/tests/unit/bsize_test.py
    /home/travis/build/linux-system-roles/storage/library/blockdev_info.py
    /home/travis/build/linux-system-roles/storage/library/resolve_blockdev.py
    /home/travis/build/linux-system-roles/storage/library/bsize.py
    /home/travis/build/linux-system-roles/storage/library/blivet.py
    /home/travis/build/linux-system-roles/storage/library/lvm_gensym.py
    /home/travis/build/linux-system-roles/storage/library/find_unused_disk.py
    /home/travis/build/linux-system-roles/storage/module_utils/size.py
Using config file /home/travis/build/linux-system-roles/storage/pylintrc
************* Module blivet
E:223,43: Instance of 'BlivetBase' has no '_volume' member (no-member)
E:250,54: Instance of 'BlivetBase' has no '_get_format' member (no-member)
E:580,96: Undefined variable 'raid_name' (undefined-variable)
ERROR: InvocationError for command /bin/bash .travis/runpylint.sh --errors-only (exited with code 2)
```
in the Python 2.7 run. Please fix this.

PR #125, opened 2020-07-07
Updated version of #64

dwlehman:
This adds fixes for the luks resize bug.


pcahyna:
Note, the fix "Don't fail to resize a non-existent device" (b23d4e8) is tested by `tests_luks.yml`.

PR #126, opened 2020-07-07
TLS/crypto param and key naming consistency

richm:
changed `encryption_key_file` to `encryption_key`
changed `encryption_passphrase` to `encryption_password`

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,14 @@ __WARNING__: Toggling encryption for a pool is a destructive operation, meaning
              the pool itself will be removed as part of the process of
              adding/removing the encryption layer.
 
-##### `encryption_passphrase`
-This string specifies a passphrase used to unlock/open the LUKS volume(s).
+##### `encryption_password`
+This string specifies a password or passphrase used to unlock/open the LUKS volume(s).
 
-##### `encryption_key_file`
-This string specifies the full path to the key file used to unlock the LUKS volume(s).
+##### `encryption_key`
+This string specifies the full path to the key file on the managed nodes used to unlock
+the LUKS volume(s).  It is the responsibility of the user of this role to securely copy
+this file to the managed nodes, or otherwise ensure that the file is on the managed
+nodes.
 
 ##### `encryption_cipher`
 This string specifies a non-default cipher to be used by LUKS.
@@ -122,11 +125,14 @@ __WARNING__: Toggling encryption for a volume is a destructive operation, meanin
              all data on that volume will be removed as part of the process of
              adding/removing the encryption layer.
 
-##### `encryption_passphrase`
-This string specifies a passphrase used to unlock/open the LUKS volume.
+##### `encryption_password`
+This string specifies a password or passphrase used to unlock/open the LUKS volume.
 
-##### `encryption_key_file`
-This string specifies the full path to the key file used to unlock the LUKS volume.
+##### `encryption_key`
+This string specifies the full path to the key file on the managed nodes used to unlock
+the LUKS volume(s).  It is the responsibility of the user of this role to securely copy
+this file to the managed nodes, or otherwise ensure that the file is on the managed
+nodes.
 
 ##### `encryption_cipher`
 This string specifies a non-default cipher to be used by LUKS.
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -10,8 +10,8 @@ storage_pool_defaults:
   type: lvm
 
   encryption: false
-  encryption_passphrase: null
-  encryption_key_file: null
+  encryption_password: null
+  encryption_key: null
   encryption_cipher: null
   encryption_key_size: null
   encryption_luks_version: null
@@ -45,8 +45,8 @@ storage_volume_defaults:
   raid_metadata_version: null
 
   encryption: false
-  encryption_passphrase: null
-  encryption_key_file: null
+  encryption_password: null
+  encryption_key: null
   encryption_cipher: null
   encryption_key_size: null
   encryption_luks_version: null
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -179,11 +179,11 @@ class BlivetBase(object):
                                                   cipher=self._spec_dict.get('encryption_cipher'),
                                                   key_size=self._spec_dict.get('encryption_key_size'),
                                                   luks_version=self._spec_dict.get('encryption_luks_version'),
-                                                  passphrase=self._spec_dict.get('encryption_passphrase') or None,
-                                                  key_file=self._spec_dict.get('encryption_key_file') or None))
+                                                  passphrase=self._spec_dict.get('encryption_password') or None,
+                                                  key_file=self._spec_dict.get('encryption_key') or None))
 
             if not device.format.has_key:
-                raise BlivetAnsibleError("encrypted %s '%s' missing key/passphrase" % (self._type, self._spec_dict['name']))
+                raise BlivetAnsibleError("encrypted %s '%s' missing key/password" % (self._type, self._spec_dict['name']))
 
             luks_device = devices.LUKSDevice(luks_name,
                                              fmt=fmt,
@@ -309,12 +309,12 @@ class BlivetVolume(BlivetBase):
 
         if device.format.type == 'luks':
             # XXX If we have no key we will always re-encrypt.
-            device.format._key_file = self._volume.get('encryption_key_file')
-            device.format.passphrase = self._volume.get('encryption_passphrase')
+            device.format._key_file = self._volume.get('encryption_key')
+            device.format.passphrase = self._volume.get('encryption_password')
 
             # set up the original format as well since it'll get used for processing
-            device.original_format._key_file = self._volume.get('encryption_key_file')
-            device.original_format.passphrase = self._volume.get('encryption_passphrase')
+            device.original_format._key_file = self._volume.get('encryption_key')
+            device.original_format.passphrase = self._volume.get('encryption_password')
             if device.isleaf:
                 self._blivet.populate()
 
--- a/tests/test-verify-volume-encryption.yml
+++ b/tests/test-verify-volume-encryption.yml
@@ -60,7 +60,7 @@
 - set_fact:
     _storage_test_expected_crypttab_entries: "{{ (storage_test_volume.encryption and _storage_test_volume_present)|ternary(1, 0) }}"
     _storage_test_crypttab_entries: "{{ storage_test_crypttab.stdout_lines|map('regex_search', '^' + storage_test_volume._device|basename + ' .*$')|select('string')|list }}"
-    _storage_test_expected_crypttab_key_file: "{{ storage_test_volume.encryption_key_file or '-' }}"
+    _storage_test_expected_crypttab_key_file: "{{ storage_test_volume.encryption_key or '-' }}"
 
 - name: Check for /etc/crypttab entry
   assert:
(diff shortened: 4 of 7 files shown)


defaults/main.yml
```diff
@@ -10,10 +10,10 @@ storage_pool_defaults:
   type: lvm
 
   encryption: false
-  encryption_passphrase: null
-  encryption_key_file: null
+  private_passphrase: null
+  private_key: null
```
tyll, line 14: "private" does not seem to make sense in this context:

1) passphrases are always private and it is not the passphrase for a private key
2) the key is not an assymmetric private key but just a (ideally randomly generated) binary blob

To group things it might make more sense to do:

For disabled encryption
```yaml
encryption: None
```

For enabled encryption:
```yaml
encryption:
    passphrase: ...
    key: ...
```

defaults/main.yml
```diff
@@ -10,10 +10,10 @@ storage_pool_defaults:
...
   encryption_cipher: null
-  encryption_key_size: null
+  private_key_size: null
```
tyll, line 16: I am pretty sure that this key size corresponds with the LUKS master key size and not the binary blob specified in `private_key`.

tyll: Basically LUKS uses one or more passphrases/binary blobs that are used to encrypt the LUKS master key. The result is stored in key slots. The master key is what is actually used to encrypt the blocks on the storage device. Not sure which details changed with LUKS2 but I guess the general concept stayed the same.

defaults/main.yml
richm: not changing `encryption` to be a `dict` value as that would involve changing more code

however, using `key` instead of `encryption_key_file` and using `password` instead of `encryption_passphrase`

defaults/main.yml
richm: reverted

nhosoi:
I guess this is not a big issue but there is no test cases to set `encryption_key_file` (now being replaced with `key`) in the test playbooks?  If it's easy, it'd be nice to add a test case to set the parameter.

richm:
@dwlehman @tyll ok?

defaults/main.yml
```diff
@@ -10,8 +10,8 @@ storage_pool_defaults:
   type: lvm
 
   encryption: false
-  encryption_passphrase: null
-  encryption_key_file: null
+  password: null
```
tyll, line 13: To me it seems that the `encryption_` prefix should be treated as a grouping mechanism like using a dictionary `encryption: ...`, and therefore `encryption_password` and `encryption_key` seem to make more sense here, for me.

dwlehman:
There are cases in both `tests_luks.yml` and `tests_luks_pool.yml` that use the `encryption_key_file` parameter.

With what are we trying to achieve a consistency?

defaults/main.yml
dwlehman: That is correct. We use prefixes for a sort of namespacing since there are so many parameters.

dwlehman:
I was wrong about the key file tests -- I'm just merging them today in #115.

richm:
ok - thanks @dwlehman and @tyll - I'd like to give @pcahyna a chance to review this before merging it.

richm:
@pcahyna - ok to merge?

richm:
@pcahyna - ok to merge?

richm:
@pcahyna - ok to merge?

richm:
@dwlehman question about `encryption_key` - This is the name of a file on the managed node - how does the file get to the managed node?

richm:
> @dwlehman question about `encryption_key` - This is the name of a file on the managed node - how does the file get to the managed node?

In the tests such as `tests_luks.yml` a tempfile is created on the managed node for the `encryption_key` - does this mean the user is expected to have copied these files to the managed node first before running the storage role?  Is this a common procedure when setting up LUKS?

dwlehman:
> > @dwlehman question about `encryption_key` - This is the name of a file on the managed node - how does the file get to the managed node?
> 
> In the tests such as `tests_luks.yml` a tempfile is created on the managed node for the `encryption_key` - does this mean the user is expected to have copied these files to the managed node first before running the storage role? Is this a common procedure when setting up LUKS?

Yes, the distribution of keys is left for the user to do however they want to. The only prior experience I have with this is kickstart, where it is the user's responsibility to get the keys into place. I have no information about what users typically do, whether there is an accepted/common practice or whether it varies from one environment to the next.

README.md
```diff
@@ -45,10 +45,10 @@ __WARNING__: Toggling encryption for a pool is a destructive operation, meaning
...
              adding/removing the encryption layer.
 
-##### `encryption_passphrase`
-This string specifies a passphrase used to unlock/open the LUKS volume(s).
+##### `encryption_password`
+This string specifies a password or passphrase used to unlock/open the LUKS volume(s).
 
-##### `encryption_key_file`
+##### `encryption_key`
 This string specifies the full path to the key file used to unlock the LUKS volume(s).
```
pcahyna, line 52: "the full path on the managed host"

README.md
```diff
@@ -122,10 +122,10 @@ __WARNING__: Toggling encryption for a volume is a destructive operation, meanin
              all data on that volume will be removed as part of the process of
              adding/removing the encryption layer.
 
-##### `encryption_passphrase`
-This string specifies a passphrase used to unlock/open the LUKS volume.
+##### `encryption_password`
+This string specifies a password or passphrase used to unlock/open the LUKS volume.
 
-##### `encryption_key_file`
```
pcahyna, line 128: "the full path on the managed host"

richm:
> Yes, the distribution of keys is left for the user to do however they want to

Ok.  I added a note to `encryption_key` in the README.md.  @dwlehman @pcahyna please check https://github.com/linux-system-roles/storage/commit/cbe7bbca70dbe8e8646f5fb575ad3cf97d298559

pcahyna:
I think this can be solved in a later release by adding a `encryption_key_src` like in nbde_client, if users desire it.
My only comments are about documentation: what is the relation between `encryption_password` and `encryption_key`? @dwlehman is `encryption_password` equivalent to the content of `encryption_key`? Similar to https://github.com/linux-system-roles/nbde_client/pull/11/files#r469968067 . It should be also clarified that this path refers to the managed host.
Note: #130 will need updating.

richm:
> I think this can be solved in a later release by adding a `encryption_key_src` like in nbde_client, if users desire it.

+1 - it might not be important if the typical LUKS use case is that the provisioning system ensures the key files are on the managed nodes.

> My only comments are about documentation: what is the relation between `encryption_password` and `encryption_key`? @dwlehman is `encryption_password` equivalent to the content of `encryption_key`? Similar to https://github.com/linux-system-roles/nbde_client/pull/11/files#r469968067 . 

Do we need to explicitly describe the relationship between `encryption_password` and `encryption_key`?

> It should be also clarified that this path refers to the managed host.

https://github.com/linux-system-roles/storage/commit/cbe7bbca70dbe8e8646f5fb575ad3cf97d298559

> Note: #130 will need updating.

. . . or I'll have to rebase this PR if #130 merges first . . .

dwlehman:
> I think this can be solved in a later release by adding a `encryption_key_src` like in nbde_client, if users desire it.
> My only comments are about documentation: what is the relation between `encryption_password` and `encryption_key`? @dwlehman is `encryption_password` equivalent to the content of `encryption_key`? Similar to https://github.com/linux-system-roles/nbde_client/pull/11/files#r469968067 . It should be also clarified that this path refers to the managed host.
> Note: #130 will need updating.

My understanding is that, typically, a key file would contain binary data, eg: output from /dev/urandom or similar.

pcahyna:
> Do we need to explicitly describe the relationship between `encryption_password` and `encryption_key`?

What about: "Providing `encryption_key` is equivalent to providing the content of this file as `encryption_password`" ?

richm:
squashed and rebased

richm:
@dwlehman if it's ok with you, I'll merge

richm:
oops - missed https://github.com/linux-system-roles/storage/pull/126#issuecomment-673579745

dwlehman:
Ref: https://access.redhat.com/solutions/230993

pcahyna:
> > I think this can be solved in a later release by adding a `encryption_key_src` like in nbde_client, if users desire it.
> > My only comments are about documentation: what is the relation between `encryption_password` and `encryption_key`? @dwlehman is `encryption_password` equivalent to the content of `encryption_key`? Similar to https://github.com/linux-system-roles/nbde_client/pull/11/files#r469968067 . It should be also clarified that this path refers to the managed host.
> > Note: #130 will need updating.
> 
> My understanding is that, typically, a key file would contain binary data, eg: output from /dev/urandom or similar.

I see. A string value is utf-8 encoded, if I understand correctly https://docs.ansible.com/ansible/latest/dev_guide/developing_python_3.html#module-string-strategy-native-string , so one can not provide arbitrary data in a string parameter, but in the special case of an ASCII key, `password` is equivalent to the content of `key`, correct?

dwlehman:
> > > I think this can be solved in a later release by adding a `encryption_key_src` like in nbde_client, if users desire it.
> > > My only comments are about documentation: what is the relation between `encryption_password` and `encryption_key`? @dwlehman is `encryption_password` equivalent to the content of `encryption_key`? Similar to https://github.com/linux-system-roles/nbde_client/pull/11/files#r469968067 . It should be also clarified that this path refers to the managed host.
> > > Note: #130 will need updating.
> > 
> > 
> > My understanding is that, typically, a key file would contain binary data, eg: output from /dev/urandom or similar.
> 
> I see. A string value is utf-8 encoded, if I understand correctly https://docs.ansible.com/ansible/latest/dev_guide/developing_python_3.html#module-string-strategy-native-string , so one can not provide arbitrary data in a string parameter, but in the special case of an ASCII key, `password` is equivalent to the content of `key`, correct?

In that specific case, yes.

dwlehman:
lgtm as long as you don't define the key file content as being necessarily the same as a password.

richm:
> but in the special case of an ASCII key, password is equivalent to the content of key, correct

Do I need to document something like this in the README.md then?
"If the password/passphrase is a UTF-8 encoded string, you can use either `encryption_key` or `encryption_password`.  If not, you must use `encryption_key`."?

dwlehman:
It's not clear to me that we need to teach users how to use LUKS. We certainly don't do the same for any of the other layers or options configurable via this role.

richm:
> It's not clear to me that we need to teach users how to use LUKS. We certainly don't do the same for any of the other layers or options configurable via this role.

Ok.  So is the text I have in the README.md in this PR sufficient?

richm:
> > It's not clear to me that we need to teach users how to use LUKS. We certainly don't do the same for any of the other layers or options configurable via this role.
> 
> Ok. So is the text I have in the README.md in this PR sufficient?

No, to answer my own question - based on the discussion above, it is too difficult to succinctly explain the relationship between `encryption_password` and `encryption_key`.  Therefore, I propose this text from the latest commit: https://github.com/linux-system-roles/storage/pull/126/commits/cf64439be70e86b63ce9baecca403459a7ded1ad
```
##### `encryption_key`
This string specifies the full path to the key file on the managed nodes used to unlock
the LUKS volume(s).  It is the responsibility of the user of this role to securely copy
this file to the managed nodes, or otherwise ensure that the file is on the managed
nodes.
```

pcahyna:
Note that `encryption_key` is documented twice.

richm:
> Note that `encryption_key` is documented twice.

Good catch.  Updated in https://github.com/linux-system-roles/storage/pull/126/commits/d0fbece9cc6f04d1c9429028ab056580a9e39de5

Ready to go?

richm:
> lgtm

@dwlehman ok?

dwlehman:
looks good

PR #135, opened 2020-07-27
Improved RAID tests

japokorn:
- RAID tests now correctly acquire value to get information about RAID
- This fixes workaround that used partially hardcoded value

### DIFF
--- a/tests/test-verify-pool-members.yml
+++ b/tests/test-verify-pool-members.yml
@@ -47,6 +47,9 @@
     loop_var: pv
   when: storage_test_pool.type == 'lvm'
 
+- name: Check raid
+  include_tasks: verify-pool-md.yml
+
 - name: Check member encryption
   include_tasks: verify-pool-members-encryption.yml
 
--- a/tests/test-verify-pool.yml
+++ b/tests/test-verify-pool.yml
@@ -3,7 +3,7 @@
 # Verify the pool configuration.
 #
 - set_fact:
-    _storage_pool_tests: ['members', 'md']
+    _storage_pool_tests: ['members']
   # future:
   #   name
   #   type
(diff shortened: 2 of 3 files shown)


japokorn:
Following pylint issues will be resolved by #137 
```
E:223,43: Instance of 'BlivetBase' has no '_volume' member (no-member)
E:250,54: Instance of 'BlivetBase' has no '_get_format' member (no-member)
E:582,96: Undefined variable 'raid_name' (undefined-variable)
```

library/blivet.py
```diff
@@ -755,12 +763,11 @@ def _create_members(self):
...
 
         if self._is_raid:
-            raid_name = "%s-1" % self._pool['name']
-
-            raid_array = self._new_mdarray(members, raid_name=raid_name)
+            raid_array = self._new_mdarray(members, raid_name=self._pool["raid_name"])
 
             self._blivet.create_device(raid_array)
             result = [raid_array]
+            self._pool['raid_name'] = raid_array.name
```
pcahyna, line 770: is it expected that you first use `self._pool["raid_name"]` and then you set it?

pcahyna:
@dwlehman does it look correct?

tests/test-verify-pool-md.yml
```diff
@@ -3,9 +3,8 @@
   # Only when pool is raid
 - name: check RAID options
   block:
-
     - name: get information about RAID
-      command: "mdadm --detail /dev/md/{{ storage_test_pool.name }}-1"
+      command: "mdadm --detail /dev/md/{{ storage_test_pool.raid_name }}"
```
dwlehman, line 7: We already have code to list and iterate over the PVs in `tests/test-verify-pool-members.yml`.  Also see `tests/test-verify-pool-members-encryption.yml`. Wouldn't it be simpler to use this model to validate the raid settings?

japokorn: I have moved the code so it is called from `test-verify-pool-members.yml`

PR #136, opened 2020-07-27
Check Volume Size in Tests

dwlehman:
This is pretty important in general, and particularly so for resize.

A keen eye may notice the commit titled "Validate volumes within test pools". Yes, that's right: Until now we have not been validating the volumes within a pool at all. I am more astonished than you, I can assure you. Also, more embarrassed.

### DIFF
--- a/tests/test-verify-pool-volumes.yml
+++ b/tests/test-verify-pool-volumes.yml
@@ -1,0 +1,7 @@
+---
+
+- name: verify the volumes
+  include_tasks: "test-verify-volume.yml"
+  loop: "{{ storage_test_pool.volumes }}"
+  loop_control:
+    loop_var: storage_test_volume
--- a/tests/test-verify-volume-size.yml
+++ b/tests/test-verify-volume-size.yml
@@ -1,0 +1,24 @@
+---
+
+- name: parse the actual size of the volume
+  bsize:
+    size: "{{ storage_test_blkinfo.info[storage_test_volume._device].size }}"
+  register: storage_test_actual_size
+  when: _storage_test_volume_present and storage_test_volume.type not in ('partition', 'disk')
+
+- name: parse the requested size of the volume
+  bsize:
+    size: "{{ storage_test_volume.size }}"
+  register: storage_test_requested_size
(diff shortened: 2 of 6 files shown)


tests/test-verify-volume-size.yml
```diff
@@ -0,0 +1,25 @@
...
+
+- debug:
+    var: storage_test_actual_size
+
+- debug:
+    var: storage_test_requested_size
+
+- assert:
+    that:
+      "{{ storage_test_actual_size == storage_test_requested_size }}"
```
pcahyna, line 23: Jinja2 template markers are unneded in `that:` : https://docs.ansible.com/ansible/latest/modules/assert_module.html

tests/test-verify-pool-volumes.yml
```diff
@@ -0,0 +1,9 @@
+---
+
+- name: verify the volumes
+  include_tasks: "test-verify-volume.yml"
+  loop: "{{ storage_test_pool.volumes }}"
+  loop_control:
+    loop_var: storage_test_volume
+  when: storage_test_pool.volumes | length > 0
```
pcahyna, line 8: Is this `when:` needed? Aren't loops over zero-length lists just skipped?

tests/test-verify-volume-size.yml
dwlehman: Fixed.

tests/test-verify-pool-volumes.yml
dwlehman: Yeah, shouldn't be needed.

richm: If it is possible that `storage_test_pool.volumes` is undefined, then perhaps something like
`when: storage_test_pool.volumes | d([])`
or even omit the `when` clause and use 
`loop: "{{ storage_test_pool.volumes | d([]) }}"`

pcahyna:
@yizhanglinux @gitPo1son @lilinredhat1 could one of you review the new test, please?

dwlehman:
Latest force-push was just to resolve merge conflict with master.

PR #137, opened 2020-07-29
Fix of overlooked pylint errors

japokorn:
- moved abstract method `get_format` from `BlivetPool` to `BlivetBase`
- fixed incorrect variables in error messages

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -158,6 +158,10 @@ class BlivetBase(object):
         self._spec_dict = spec_dict
         self._device = None
 
+    def _get_format(self):
+        # abstract method
+        raise NotImplementedError()
+
     def _manage_one_encryption(self, device):
         ret = device
         # Make sure to handle adjusting both existing stacks and future stacks.
@@ -218,9 +222,9 @@ class BlivetBase(object):
         if requested_actives is not None and requested_spares is not None:
             if (requested_actives + requested_spares != len(members) or
                     requested_actives < 0 or requested_spares < 0):
-                raise BlivetAnsibleError("failed to set up volume '%s': cannot create RAID "
+                raise BlivetAnsibleError("failed to set up '%s': cannot create RAID "
                                         "with %s members (%s active and %s spare)"
-                                        % (self._volume["name"], len(members),
+                                        % (self._spec_dict["name"], len(members),
                                             requested_actives, requested_spares))
 
         if requested_actives is not None:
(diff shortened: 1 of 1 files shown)


library/blivet.py
```diff
@@ -579,7 +584,7 @@ def _create(self):
             try:
                 do_partitioning(self._blivet)
             except Exception as e:
-                raise BlivetAnsibleError("failed to allocate partitions for mdraid '%s': %s" % (raid_name, str(e)))
```
pcahyna, line 582: would it make sense to replace `raid_name` by `self._volume["name"]` instead of removing it completely?

japokorn: Most of the information needed should be already included in e.

library/blivet.py
```diff
@@ -152,12 +152,17 @@ class BlivetAnsibleError(Exception):
 class BlivetBase(object):
     blivet_device_class = None
     _type = None
+    _volume = None
```
dwlehman, line 155: Use `self._spec_dict` in the base class. You only have `self._volume` in the volume subclasses and `self._pool` in the pool subclasses.

library/blivet.py
dwlehman: I agree that the volume name could be useful here. The partitioning exception raised by blivet is not likely to indicate the name of the volume being configured.

library/blivet.py
japokorn: Agree. The reason of `_volume` being used here is that some `BlivetBase` methods work with `self._volume` and pylint doesn't like it since it is undefined. Other way to fix this would be to disable pylint warnings. I think that using 'abstract variable' is slightly better solution because it handles all the future `self._volume` invocations from this class.
I should probably add some explanation into the code, though.

japokorn: I have eventually decided to disable pylint. Since dummy variable hadn't had all required attributes of the `self._volume`, pylint noticed it anyways. This also nullifies my previous argument about future invocations.

library/blivet.py
japokorn: I see. Fixed.

pcahyna:
@dwlehman does it look good now?

library/blivet.py
```diff
@@ -220,7 +224,7 @@ def _new_mdarray(self, members, raid_name=""):
                     requested_actives < 0 or requested_spares < 0):
                 raise BlivetAnsibleError("failed to set up volume '%s': cannot create RAID "
                                         "with %s members (%s active and %s spare)"
-                                        % (self._volume["name"], len(members),
+                                        % (self._volume["name"], len(members),  # pylint: disable=no-member
```
dwlehman, line 227: Is this method not used for pools? If so, it must use `self._spec_dict` instead of either `self._volume` or `self._pool`; if not, it should be in the `BlivetVolume` class instead.

japokorn: Fixed. Replaced with `self._spec_dict`. Removed `pylint: disable`.

japokorn:
Fixed: Removed `_get_format` method from `BlivetPool`.

PR #142, opened 2020-08-07
Changed size.py path

japokorn:
- moved `size.py` to `module_utils/storage`
- modified affected imports


richm:
There is a problem with the https://github.com/linux-system-roles/storage/blob/master/tests/setup_module_utils.sh script that sets up the python/ansible environment for pyunit and pylint.  The problem is that ansible already has a module_utils/storage/.  The setup script expects that there is no `module_utils/$rolename/` directory.  I suppose instead of setting up a symlink from ansible `module_utils/$rolename/` to role `module_utils/$rolename/` the script could copy the files e.g. `cp role module_utils/$rolename/*.py ansible module_utils/$rolename/`  but that might also be a problem if the file already exists.

Another suggestion - use a unique name for the role module_utils directory e.g. `module_utils/lsr_storage/` and
```
from ansible.module_utils.lsr_storage.size import Size
```
@dwlehman @pcahyna ?

richm:
hmm - well it seems that the integration tests are passing - assuming they are exercising the changed code, I guess ansible is doing something like adding `size.py` to the `ansible.module_utils.storage` namespace, and it's ok because there is no built-in `ansible.module_utils.storage.size`

so - how do you guys want to handle this?

richm:
> so - how do you guys want to handle this?

I would prefer to use a unique directory name - something like `module_utils/lsr_storage/` - so that there is no possibility of collision.

pcahyna:
how will this work in the Collection world? Will there be a single namespace for the whole collection? Can we do this in a way that it prepares for future work on collections? @nhosoi

dwlehman:
> > so - how do you guys want to handle this?
> 
> I would prefer to use a unique directory name - something like `module_utils/lsr_storage/` - so that there is no possibility of collision.

I think this module definitely qualifies as being generically useful for storage operations/calculations, so my preference (without knowing the ansible conventions or having discussed w/ upstream) would be to include it under 'storage/'. I also like something along the lines of 'lsr_storage/', but not as much for this particular module.

pcahyna:
It would be a longer term project and definitely separate from this PR, but if the functionality is generically useful, it could be added to human_to_bytes filter ( https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#human-to-bytes ), the actual implementation lives in `ansible.module_utils.common.text.formatters`.

dwlehman:
I didn't know that was there. It's a bit frustrating that it doesn't differentiate between KB and KiB but whatever.

It's probably worth noting that `bsize` is not used at all in the role currently. (Nor is `lvm_gensym` or `resolve_blockdev`.)

nhosoi:
> how will this work in the Collection world? Will there be a single namespace for the whole collection? Can we do this in a way that it prepares for future work on collections? @nhosoi

This is the current snapshot of the role2collection conversion tool.
https://github.com/nhosoi/auto-maintenance/blob/role2collection/lsr-role2collection.py
There's no problem supporting the both cases.
```
- from ansible.module_utils.size import Size
+ from ansible.module_utils.storage.size import Size
```

pcahyna:
@nhosoi , shouldn't the import path be collection-specific? I would expect to see `system_roles` (the name of the collection) in the import path.

nhosoi:
> @nhosoi , shouldn't the import path be collection-specific? I would expect to see `system_roles` (the name of the collection) in the import path.

@pcahyna, yes, it is taken care. (maybe, we could discuss it in the tool's pr to be opened... ;)

pcahyna:
>  yes, it is taken care

I don't understand. By whom and how?

richm:
I'll note that the network role uses `module_utils/network_lsr` - https://github.com/linux-system-roles/network/tree/main/module_utils/network_lsr - I suggest storage use `module_utils/storage_lsr` if necessary (if collections doesn't solve this problem _en passant_).

richm:
closing in favor of https://github.com/linux-system-roles/storage/pull/177

PR #143, opened 2020-08-07
sync with latest template including shellcheck

richm:
sync with latest template including shellcheck
fix shellcheck issues

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 [tox]
 envlist =
     black, pylint, flake8, yamllint
-    py{26,27,36,37,38},
+    py{26,27,36,37,38}, shellcheck
     custom
 skipsdist = true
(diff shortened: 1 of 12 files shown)


tox.ini
```diff
@@ -193,7 +199,7 @@ max-line-length = 88
 python =
   2.6: py26,coveralls,custom
   2.7: py27,coveralls,flake8,pylint,custom
-  3.6: py36,coveralls,black,yamllint,custom
+  3.6: py36,coveralls,black,yamllint,shellcheck,custom
```
dwlehman, line 202: Does this mean we only run shellcheck on python3.6? If so, why?

richm: shellcheck isn't really a python thing - we only need to run it once, no matter what the python version is - we're just using tox as a convenient task runner - same with yamllint

PR #144, opened 2020-08-08
Avoid using Jinja2 template markers in assert

yizhanglinux:
Avoid using Jinja2 template markers in assert

Fixes: linux-system-roles#127

### DIFF
--- a/tests/tests_deps.yml
+++ b/tests/tests_deps.yml
@@ -26,8 +26,8 @@
 
     - assert:
         that:
-          - "{{ blivet_output.packages }} == ['lvm2', 'xfsprogs']"
-          - "{{ not blivet_output.changed }}"
+          - blivet_output.packages == ['lvm2', 'xfsprogs']
+          - not blivet_output.changed
         msg: "unexpected required package list: {{ blivet_output.packages }}"
 
     - name: test disk and ext4 package deps
@@ -44,8 +44,8 @@
 
     - assert:
         that:
-          - "{{ blivet_output.packages}} == ['e2fsprogs']"
-          - "{{ not blivet_output.changed }}"
+          - blivet_output.packages == ['e2fsprogs']
+          - not blivet_output.changed
         msg: "unexpected required package list: {{ blivet_output.packages }}"
 
     - name: test disk and swap package deps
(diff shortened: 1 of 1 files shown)


pcahyna:
Thanks for the suggestion, but I suspect the problem lies in the blivet module itself. One should not be forced to do such conversions in the Ansible/Jinja2 code.

tests/tests_deps.yml
```diff
@@ -26,7 +26,7 @@
 
     - assert:
         that:
-          - "{{ blivet_output.packages }} == ['lvm2', 'xfsprogs']"
+          - "{{ blivet_output.packages|to_json }} == ['lvm2', 'xfsprogs']"
```
richm, line 29: Does this work?
```yaml
           - blivet_output.packages == ['lvm2', 'xfsprogs']
```
?
If it doesn't work, is it because `blivet_output.packages` is not sorted?  So then, does this work?
```yaml
           - blivet_output.packages|sort == ['lvm2', 'xfsprogs']
```

yizhanglinux: > Does this work?
> 
> ```yaml
>            - blivet_output.packages == ['lvm2', 'xfsprogs']
> ```
> 
Yeah, this format works well.


> ?
> If it doesn't work, is it because `blivet_output.packages` is not sorted? So then, does this work?
> 
> ```yaml
>            - blivet_output.packages|sort == ['lvm2', 'xfsprogs']
> ```

richm: For `when` statements and `assert.that` statements, you generally don't need quoting and `{{ }}` expansion

pcahyna:
@richm good catch! I keep forgetting that one does not need the Jinja2 template markers in assert and when.

tests/tests_deps.yml
yizhanglinux: OK, if you agree, I will cook one patch to remove all the quoting and `{{ }}` expansion in `when` statements and `assert.that` statements

pcahyna: @yizhanglinux thanks, but please start with updating this PR to remove it at this problematic location, you don't need to fix it everywhere immeadiately.

yizhanglinux: > @yizhanglinux thanks, but please start with updating this PR to remove it at this problematic location, you don't need to fix it everywhere immeadiately.

OK, will update it later.

richm:
[citest bad]

pcahyna:
This also applies to the commit message of the commit, which no longer describes it correctly.

yizhanglinux:
Close it and using #147 to track it

yizhanglinux:
> This also applies to the commit message of the commit, which no longer describes it correctly.

Thanks, I just submitted another PR with updated commit message/title/pr branch name and merged it.

PR #147, opened 2020-08-12
Avoid using Jinja2 template markers in assert

yizhanglinux:
Fixes: linux-system-roles#127


yizhanglinux:
This branch was used for changing the PR name of #144

PR #152, opened 2020-08-19
Add new test case for FS resize

gitPo1son:
1.resize to size_after more then  size_beforce
2.resize to size_after less then  size_beforce
3.resize to too-large volume size
4.resize to volume size equal disk's size
5.resize to invalid size


yizhanglinux:
[citest bad]

PR #153, opened 2020-08-20
Be smarter in choosing expected partition name.

dwlehman:
BlivetVolume._get_device_id is only used to look up pre-existing
volumes, so we don't have to try too hard to guess it by name.
We can just see if the disk has a single partition and, if so,
return the name of that partition.

Fixes: #141

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -556,7 +556,11 @@ class BlivetPartitionVolume(BlivetVolume):
         return self._device.raw_device.type == 'partition'
 
     def _get_device_id(self):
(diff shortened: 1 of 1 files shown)


dwlehman:
While an automated test would be great, I think the most important thing is to get this fixed so that customers don't see it in their environments.

yizhanglinux:
> While an automated test would be great, I think the most important thing is to get this fixed so that customers don't see it in their environments.

agree, let's merge it first.

richm:
What is the plan for this PR?  Any update?

PR #128, opened 2020-07-14
Pass -F to mke2fs for whole disks in RHEL7.

dwlehman:
Fixes #122

PR #154, opened 2020-08-20
WIP: rebased version of #128 with suggested change

pcahyna:


pcahyna:
it is WIP because I need tpo document the new option, but the code should be ready.

vars/CentOS_7.yml
```diff
@@ -6,3 +6,7 @@ blivet_package_list:
   - libblockdev-lvm
   - libblockdev-mdraid
   - libblockdev-swap
+__storage_blivet_mkfs_option_map:
```
dwlehman, line 9: Only `BlivetDiskVolume` uses this map. Shouldn't the map also have something to indicate which volume type the arguments apply to?

PR #161, opened 2020-09-14
Non-existent pool removal fix

japokorn:
- removing non-existent pool with underlaying disks no longer fails

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -307,7 +307,11 @@ class BlivetVolume(BlivetBase):
         if self._device:
             return
 
-        device = self._blivet.devicetree.resolve_device(self._get_device_id())
+        device_id = self._get_device_id()
+        if device_id is None:
+            return
+
+        device = self._blivet.devicetree.resolve_device(device_id)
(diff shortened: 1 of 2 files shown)


japokorn:
Running following playbook on a system without pool present on `/dev/vdb` crashed. However if the pool is present it works normally.

```
---
- hosts: all
  become: true
  vars:
    storage_safe_mode: false
    storage_use_partitions: true

  tasks:
    - name: test1
      include_role:
        name: linux-system-roles.storage
      vars:
        storage_pools:
          - name: foo
            disks:
             - '/dev/vdb'
            state: absent
            volumes:
              - name: test1
                mount_point: '/opt/test1'
```

The role identifies non-existent `test1` volume and tries to get information about it from its parent pool object. Pool object does not exist and Exception is raised.

dwlehman (commented):
Please add test cases covering removing a non-existing pool both with and without volumes. Otherwise it looks good.

dwlehman (approved):
Looks good. Thanks!

Issue #162, opened 2020-09-14
Missing or incorrect type for volumes in storage_volumes yields a confusing error message


dwlehman:
You get an unhandled traceback when it should be a simple, clear message about the type being incorrect/invalid.

pcahyna:
fixed in #165 right?

PR #165, opened 2020-09-16
Confusing error message fix

japokorn:
- missing or incorrect type for volumes in storage_volumes no longer yields a confusing error message

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -1174,7 +1174,10 @@ def run_module():
     actions = list()
 
     if module.params['packages_only']:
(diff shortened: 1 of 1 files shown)


japokorn:
This fixes #162

dwlehman:
What error message is produced with this patch? I was expecting new code in `_get_blivet_volume` to identify this case and raise an exception.

japokorn:
@dwlehman `_get_blivet_volume` already produces `BlivetAnsibleError`. The missing part in this case is exception handling by module (`module.fail_json`) during "packages_only" run.

dwlehman (approved):
Looks good to me. Thanks for the explanation.

PR #168, opened 2020-09-22
Disallow toggling encryption in safe mode

pcahyna:
Toggling encryption is (currently) a destructive operation, and this fact is quite nonintuitive and possibly dangerous. Let it be covered by the storage_safe_mode flag, just as with other operations that cause data loss (except for explicit removal).

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -167,11 +167,16 @@ class BlivetBase(object):
         raise NotImplementedError()
 
     def _manage_one_encryption(self, device):
+        global safe_mode
         ret = device
         # Make sure to handle adjusting both existing stacks and future stacks.
         if device == device.raw_device and self._spec_dict['encryption']:
             # add luks
             luks_name = "luks-%s" % device._name
+            if safe_mode and (device.original_format.type is not None or
+                              device.original_format.name != get_format(None).name):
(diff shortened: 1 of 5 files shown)


dwlehman (approved):
It looks good to me. My only comment is that we should be validating the value of `blivet_output.changed` to make sure it is `False` when we abort due to `storage_safe_mode`. It doesn't necessarily have to be in this patch set, though, I guess.

pcahyna:
> It looks good to me. My only comment is that we should be validating the value of `blivet_output.changed` to make sure it is `False` when we abort due to `storage_safe_mode`. It doesn't necessarily have to be in this patch set, though, I guess.

All the new tests have it, don't they? I just pushed a commit to improve formatting of this part, no functional change intended.

pcahyna:
Actually, some of tests don't have it. Let me fix it.

pcahyna:
I added those checks, which allowed me to find and fix a small error in tests: one pool name was given as `foobar` and not `foo`, here: https://github.com/linux-system-roles/storage/commit/8c091bd4828392216798dcda6575977aa53294a4#diff-ede83c510a6055640001e6b500925559R334 . I also added tests for partition pools.

yizhanglinux:
@guazhangRH here is the luks fix MR

Issue #163, opened 2020-09-14
raid volumes do not remove existing data from member disks as needed before creation


dwlehman:
It should use `self._blivet.devicetree.recursive_remove` like `BlivetPool._create_members` does.

AbdelAzizMohamedMousa:
It is generally recommended to remove existing data from member disks before creating a RAID volume to avoid potential data loss or corruption. The recommended way to do this in Blivet is to use the recursive_remove() method of the DeviceTree object.

In the code snippet you provided, it seems that the recursive_remove() method is not used to remove existing data from member disks before creating a RAID volume. To ensure that existing data is removed, you can modify the code to use recursive_remove() as follows: 
`def create(self):
    for device in self.devices:
        # remove existing data from member disks
        self._blivet.devicetree.recursive_remove(device)
        
    # create RAID volume
    self._blivet.devicetree.add_device(self._blivet.dmraid_factory(*self.devices))
`
By calling recursive_remove() on each device before creating the RAID volume, you can ensure that any existing data on the member disks is removed.

japokorn:
Fixed by #169.

PR #169, opened 2020-09-22
fix: RAID volume pre cleanup 

japokorn:
- raid volumes now remove existing data from member disks as needed before creation

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -1002,8 +1002,13 @@ class BlivetMDRaidVolume(BlivetVolume):
         if self._device:
             return
 
-        if safe_mode:
-            raise BlivetAnsibleError("cannot create new RAID in safe mode")
+        for spec in self._volume["disks"]:
+            disk = self._blivet.devicetree.resolve_device(spec)
+            if not disk.isleaf or disk.format.type is not None:
+                if safe_mode and (disk.format.type is not None or disk.format.name != get_format(None).name):
+                    raise BlivetAnsibleError("cannot remove existing formatting and/or devices on disk '%s' in safe mode" % disk.name)
+                else:
+                    self._blivet.devicetree.recursive_remove(disk)
 
         # begin creating the devices
         members = self._create_raid_members(self._volume["disks"])
--- a/tests/tests_raid_volume_cleanup.yml
+++ b/tests/tests_raid_volume_cleanup.yml
@@ -1,0 +1,102 @@
+---
+- name: Test RAID cleanup
+  hosts: all
+  become: true
+  vars:
+    storage_safe_mode: false
+    storage_use_partitions: true
+    mount_location1: '/opt/test1'
+    mount_location2: '/opt/test2'
+    volume1_size: '5g'
+    volume2_size: '4g'
+
+  tasks:
+    - name: Run the role
+      include_role:
+        name: linux-system-roles.storage
+
+    - name: Mark tasks to be skipped
+      set_fact:
+        storage_skip_checks:
+          - blivet_available
+          - packages_installed
+          - service_facts
+
+    - name: Get unused disks
+      include_tasks: get_unused_disk.yml
+      vars:
+        max_return: 3
+        disks_needed: 3
+
+    - name: Create two LVM logical volumes under volume group 'foo'
+      include_role:
+        name: linux-system-roles.storage
+      vars:
+        storage_pools:
+          - name: foo
+            disks: "{{ unused_disks }}"
+            volumes:
+              - name: test1
+                size: "{{ volume1_size }}"
+                mount_point: "{{ mount_location1 }}"
+              - name: test2
+                size: "{{ volume2_size }}"
+                mount_point: "{{ mount_location2 }}"
+
+    - name: Enable safe mode
+      set_fact:
+        storage_safe_mode: true
+
+    - name: >-
+        Try to overwrite existing device with raid volume
+        and safe mode on (expect failure)
+      include_tasks: verify-role-failed.yml
+      vars:
+        __storage_failed_regex: cannot remove existing formatting.*in safe mode
+        __storage_failed_msg: >-
+          Unexpected behavior when overwriting existing device with RAID volume
+        __storage_failed_params:
+          storage_volumes:
+            - name: test1
+              type: raid
+              raid_level: "raid1"
+              raid_device_count: 2
+              raid_spare_count: 1
+              disks: "{{ unused_disks }}"
+              mount_point: "{{ mount_location1 }}"
+              state: present
+
+    - name: Disable safe mode
+      set_fact:
+        storage_safe_mode: false
+
+    - name: Create a RAID0 device mounted on "{{ mount_location1 }}"
+      include_role:
+        name: linux-system-roles.storage
+      vars:
+        storage_volumes:
+          - name: test1
+            type: raid
+            raid_level: "raid1"
+            raid_device_count: 2
+            raid_spare_count: 1
+            disks: "{{ unused_disks }}"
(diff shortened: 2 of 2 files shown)


japokorn:
Fixes #163

pcahyna:
is there a test?

library/blivet.py
```diff
@@ -611,6 +611,11 @@ def _create(self):
         if safe_mode:
             raise BlivetAnsibleError("cannot create new RAID in safe mode")
```
pcahyna, line 612: I can't comment on the change itself, but this line above does not look right. Does it mean that every RAID creation is considered unsafe and users must set safe_mode off for it to be allowed, even if the RAID is to be created on an empty disk set?

japokorn: Fixed. Thanks for noticing.

pcahyna: We must not fall into the opposite problem of allowing to destroy existing data in safe mode though. See #168 .

library/blivet.py
```diff
@@ -608,8 +608,13 @@ def _create(self):
         if self._device:
             return
 
-        if safe_mode:
-            raise BlivetAnsibleError("cannot create new RAID in safe mode")
+        for spec in self._volume["disks"]:
+            disk = self._blivet.devicetree.resolve_device(spec)
+            if not disk.isleaf or disk.format.type is not None:
+                if safe_mode:
+                    raise BlivetAnsibleError("cannot create new RAID in safe mode")
```
dwlehman, line 615: It would be better to use a slight variation on the error raised from `BlivetPool._create_members`.

dwlehman (commented):
Looks good otherwise.

library/blivet.py
japokorn: Fixed: Used slightly modified version of error from `BlivetPool._create_members`.

library/blivet.py
```diff
@@ -623,8 +623,13 @@ def _create(self):
         if self._device:
             return
 
-        if safe_mode:
-            raise BlivetAnsibleError("cannot create new RAID in safe mode")
+        for spec in self._volume["disks"]:
+            disk = self._blivet.devicetree.resolve_device(spec)
+            if not disk.isleaf or disk.format.type is not None:
```
dwlehman, line 628: The check for `safe_mode` has to be a bit more careful, like the ones in `BlivetVolume._reformat` and `BlivetBase._manage_one_encryption`. I see that `BlivetPool._create_members` also needs to check for `device.original_format.name != get_format(None).name` (to catch formatting reported by blkid but not recognized/handled by blivet).

dwlehman (changes requested):
My suggestion would be to make the change where is applies to what you're doing here and maybe open a separate pull request to address the issue in `BlivetPool._create_members`.

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

library/blivet.py
japokorn: Fixed.

japokorn:
Rebased. This should fix the issue with failing test.

richm:
[citest bad]

richm:
[citest pending]

richm:
Do we still need this?

richm:
[citest pending]

japokorn:
Nilfs2 is now used only on systems that provide it. Also rebased to latest version.

richm:
[citest bad]

japokorn:
[citest]

japokorn:
[citest bad]

tests/tests_raid_volume_cleanup.yml
```diff
@@ -0,0 +1,105 @@
+---
+- hosts: all
```
richm, line 2:
```suggestion
- name: Test RAID cleanup
  hosts: all
```

tests/tests_raid_volume_cleanup.yml
```diff
@@ -0,0 +1,105 @@
...
+  vars:
+    storage_safe_mode: false
+    storage_use_partitions: true
+    mount_location1: '/opt/test1'
+    mount_location2: '/opt/test2'
+    volume1_size: '5g'
+    volume2_size: '4g'
+
+  tasks:
+    - include_role:
```
richm, line 13:
```suggestion
    - name: Call the storage role
      include_role:
```

tests/tests_raid_volume_cleanup.yml
```diff
@@ -0,0 +1,105 @@
...
+        name: linux-system-roles.storage
+
+    - name: Mark tasks to be skipped
+      set_fact:
+        storage_skip_checks:
+          - blivet_available
+          - packages_installed
+          - service_facts
+
+    - include_tasks: get_unused_disk.yml
```
richm, line 23:
```suggestion
    - name: Get unused disks
      include_tasks: get_unused_disk.yml
```

tests/tests_raid_volume_cleanup.yml
```diff
@@ -0,0 +1,105 @@
...
+                mount_point: "{{ mount_location1 }}"
+              - name: test2
+                size: "{{ volume2_size }}"
+                mount_point: "{{ mount_location2 }}"
+
+    - name: Enable safe mode
+      set_fact:
+        storage_safe_mode: true
+
+    - name: Try to overwrite existing device with raid volume and safe mode on (expect failure)
```
richm, line 47:
```suggestion
    - name: >-
        Try to overwrite existing device with raid volume and safe mode on
        (expect failure)
```

tests/tests_raid_volume_cleanup.yml
```diff
@@ -0,0 +1,105 @@
...
+              - name: test1
+                type: raid
+                raid_level: "raid1"
+                raid_device_count: 2
+                raid_spare_count: 1
+                disks: "{{ unused_disks }}"
+                mount_point: "{{ mount_location1 }}"
+                state: present
+
+        - name: unreachable task
```
richm, line 63:
```suggestion
        - name: Unreachable task
```

tests/tests_raid_volume_cleanup.yml
```diff
@@ -0,0 +1,105 @@
...
+                size: "{{ volume2_size }}"
+                mount_point: "{{ mount_location2 }}"
+
+    - name: Enable safe mode
+      set_fact:
+        storage_safe_mode: true
+
+    - name: Try to overwrite existing device with raid volume and safe mode on (expect failure)
+      block:
+        - name: Create a RAID0 device mounted on "{{ mount_location1 }}"
```
richm, line 49: For checking that the role raises an error, and that the error message is the one that you expect, please consider using verify-role-failed.yml as in https://github.com/linux-system-roles/storage/blob/main/tests/tests_misc.yml#L66
One of the issues we have had in the past with testing for storage role failures was that the role would fail in the wrong way i.e. because the inputs were not correct, or there was some other blivet error, but the test would report that the role failed correctly because it was not checking for the error message - using verify-role-failed.yml makes it easy to verify that the role failed *AND* that the error message is the correct one for the failure condition.

tests/tests_raid_volume_cleanup.yml
```diff
@@ -0,0 +1,105 @@
...
+        storage_volumes:
+          - name: test1
+            type: raid
+            raid_level: "raid1"
+            raid_device_count: 2
+            raid_spare_count: 1
+            disks: "{{ unused_disks }}"
+            mount_point: "{{ mount_location1 }}"
+            state: absent
+
```
richm, line 105:
```suggestion
```

richm:
testing now

richm (approved):

PR #172, opened 2020-09-28
WIP - add a special setup for citool/openstack environments

richm:
Setup that uses targetcli to create disk devices when running
in the citool/openstack environment.


richm:
@ukulekek @pcahyna ptal - this makes citool tests pass for the storage role when using `guest-setup --playbooks setup_for_citool.yml`

PR #124, opened 2020-07-06
Don't try to set up mounts with invalid mount points.

dwlehman:
Fixes #111

PR #179, opened 2020-12-01
Fake pr for debugging https://github.com/linux-system-roles/test-harmess/issues/124

nhosoi:
CI tests with collections option fail with "__spec__ is None" if the role has module_utils #124


nhosoi:
Good results - all staging (citool-staging is not counted here) passed.

PR #183, opened 2020-12-17
use new tox-lsr plugin

richm:
Remove the shared scripts and instead use the new tox-lsr plugin.


dwlehman:
Are any of those CI failures unexpected? The python (3.6) one, for example, seems to have succeeded, but exited with an error code. I have no idea if that's specific to these changes.

richm:
> Are any of those CI failures unexpected? The python (3.6) one, for example, seems to have succeeded, but exited with an error code. I have no idea if that's specific to these changes.

They are expected.  `ERROR: Cannot install ansible-lint==4.3.5 and ansible==2.7.* because these package versions have conflicting dependencies.`  - in a subsequent PR we will drop support for ansible 2.7 in CI.  We are also going to
- drop support for centos6 in molecule
- drop support for linting using molecule and use e.g. tox -e ansible-lint
- add centos8 support in molecule

dwlehman (approved):
Looks okay to me.

PR #184, opened 2020-12-23
use tox for ansible-lint instead of molecule

richm:
Deprecate the use of molecule for linting.  Use tox for ansible-lint.
The next step will be to remove the ansible-lint dependency from
the molecule environments.


richm:
please review
The CI failure is known and expected and will be resolved by an upcoming commit

dwlehman (approved):

PR #190, opened 2021-01-21
Refined safe_mode condition in create_members

japokorn:
- `safe_mode` condition in `Blivet.Pool._create_members` should now catch
  formatting reported by `blkid` but not being recognized/handled by blivet


jharuda:
[citest]

japokorn:
[citest]

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

nhosoi:
[citest]

library/blivet.py
```diff
@@ -785,7 +785,8 @@ def _create_members(self):
 
         for disk in self._disks:
             if not disk.isleaf or disk.format.type is not None:
-                if safe_mode:
+                #if safe_mode:
```
nhosoi, line 788: Could you remove this commented out line?

richm:
> @japokorn is there a test that shows the problematic behavior and verifies the fix? Is it a serious issue (i.e. possible data loss)?

and please rebase

japokorn:
@pcahyna There is a possibility that existing device will not be recognized by blivet but `blkid` will be able get information about it.
This change fortifies the check in a way that `blkid` result is taken into account as well.
The same construct is used in other part of the code, this one was missed.

pcahyna:
> @pcahyna There is a possibility that existing device will not be recognized by blivet but `blkid` will be able get information about it.
> This change fortifies the check in a way that `blkid` result is taken into account as well.

@japokorn what is the storage configuration(s) that exhibits the problem? I am not familiar enough with blivet to be able to tell when an existing device will be recognized by `blkid` but not by blivet. Can you add it to tests, please?

japokorn:
@pcahyna: I really do not think that writing the test for this fix will be worth the effort. This fix should help prevent rather niché situations (new type of device was developed and blivet cannot handle it yet) with almost no impact on the code. The same condition is already used in multiple other places and is being processed in existing tests.
To actually create the test would probably require either somehow block part of blivet functionality or trick `blkid` to report non-existing device. Doing this in playbook would be extremely difficult. Writing a new python module might help but it would be time consuming. Also we do not use any modules for the testing at the moment and setting this up would take some time as well. Even with the module in place it would not be an easy task. I believe that my time will be better spent elsewhere.

richm:
> I believe that my time will be better spent elsewhere

In order to get this fix into the product, this fix will need to pass QE and Docs review - what does QE need to know to verify this fix, and what does Docs need to know to document it?

pcahyna:
@japokorn do I understand correctly that currently there are no known storage layouts that would trigger the problem and the fix merely prevents hypothetical future situations?

japokorn:
@richm, @pcahyna: Based on your comments I did some research into this issue. It is probably possible to simulate the issue by creating `nilfs2` filesystem which is currently unsupported by blivet. Unrecognized format should trigger the condition. I should be able to write the test for it.

richm:
looks like `nilfs` is not available on several platforms: https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-storage-pull-linux-system-roles_storage-190-cf86a37-rhel-x-20210609-233748/artifacts/ansible.log
```
TASK [install nilfs2 support] **************************************************
task path: /tmp/tmpafn86qt4/tests/tests_safe_mode_check.yml:21
fatal: [/cache/rhel-x.qcow2]: FAILED! => {"changed": false, "failures": ["No package nilfs-utils available."], "msg": "Failed to install some of the specified packages", "rc": 1, "results": []}
```
You might have to restrict the test to Fedora

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

richm:
[citest pending]

tests/tests_safe_mode_check.yml
```diff
@@ -0,0 +1,87 @@
+---
+- hosts: all
+  become: true
+  vars:
+    storage_safe_mode: true
+    mount_location1: '/opt/test1'
+    mount_location2: '/opt/test2'
+    volume_group_size: '10g'
+    volume1_size: '5g'
+    volume2_size: '4g'
```
richm, line 10: please add
```yaml
  tags:
    - tests::lvm
```

richm:
also, if you rebase, you can fix the tox errors

one minor nit, otherwise, lgtm - @dwlehman ?

vojtechtrefny (approved):

richm:
ok to merge?

tests/tests_safe_mode_check.yml
japokorn: Fixed.

japokorn:
[citest bad]

PR #194, opened 2021-02-08
drop localhost from tests_deps.yml

richm:
We do not need to check the deps on localhost.  This is causing
test failures in the staging environment (which will soon become
the new production environment) due to the way we test now with
ANSIBLE_ROLES_PATH environment variable instead of the tests/roles
role symlink

### DIFF
--- a/tests/tests_deps.yml
+++ b/tests/tests_deps.yml
@@ -1,5 +1,5 @@
 ---
-- hosts: all localhost
+- hosts: all
   become: true
 
   tasks:


dwlehman (approved):
Nice -- thanks.

dwlehman:
Hmm, maybe I was hasty. Does this fix anything yet?

richm:
> Hmm, maybe I was hasty. Does this fix anything yet?

Yes.  `tests_deps.yml` was consistently failing in the staging environment.  I'm not entirely sure why, but this change seems to fix that.  I don't know why `tests_deps.yml` needs to check on `localhost`, so I removed that check, and the test started working. 
The rhel-x failures are known and are fixed by checking for crypttab
The fedora-32 failure was due to a transient network repo problem

richm:
@dwlehman ok - please review

dwlehman (approved):
Looks good. I am fairly certain that `localhost` is a relic from initial testing in a vm, and completely certain it should be removed.

richm:
> Looks good. I am fairly certain that `localhost` is a relic from initial testing in a vm, and completely certain it should be removed.

ok - if you need to do localhost testing with this playbook, you can use `ansible-playbook -c local -i localhost, tests_deps.yml`

richm:
[citest bad]

PR #195, opened 2021-02-09
use molecule v3, drop v2 - use tox-lsr 2.1.2

richm:
Switch to molecule v3.  This drops support for v2.
This requires tox-lsr 2 or later.
Upgrade to tox-lsr 2.1.2 to pick up fix for collection dependency
on python six.


richm:
@dwlehman as you can see this fixes the tox issues in CI

dwlehman (approved):
Looks good -- thanks.

Issue #13, opened 2018-11-24
Insufficient space when lv size matches vg


tabowling:
Playbook fails when lv size is equal to vg size.

```
TASK [linux-system-roles.storage : Make sure LV exists] ***************************************************************************************************************************
fatal: [rhel7]: FAILED! => {"changed": false, "err": "  Volume group \"composer\" has insufficient free space (5119 extents): 5120 required.\n", "msg": "Creating logical volume 'composer' failed", "rc": 5}

[root@rhel7 ~]# vgs
  VG       #PV #LV #SN Attr   VSize   VFree  
  composer   1   0   0 wz--n- <20.00g <20.00g
  rhel       1   2   0 wz--n-  <7.00g      0 
```

Playbook used was:

```
- hosts: all
  become: yes
  become_method: sudo
  become_user: root

  vars:
    use_partitions: false
  tasks:
    - name: Configure Composer Storage
      include_role:
        name: linux-system-roles.storage
      vars:
        storage_pools:
          - name: composer
            disks: ['vdb']
            # type: lvm
            state: present
            volumes:
              - name: composer
                size: 20G
                # type: lvm
                # fs_type: xfs
                fs_label: "composer"
                mount_point: '/var/lib/lorax/composer'
```

dwlehman:
IMO this is an LVM bug, and they've agreed there should be an option to interpret size as "up to this size" for several years now, but no code has been written. We can try to handle this in the role, but it will take a bit of work (which is fraught with danger due to guesswork).

pcahyna:
Is there any bug actually? The output `<20.00g` indicates there is not exactly 20g of space, it is rounded and there is actually less space than that. I think it is better to let the role fail than let it create a smaller volume than what you asked for.

dwlehman:
The percentage-based size work (#39) will address this issue in another way, but I am also open to truncating sizes as needed up to some agreed-upon percentage to make things work seamlessly without requiring the user to predict the total space used by metadata .

When I wrote blivet many years ago I was determined to create volumes/devices the exact size requested by the user. Over the years I have gradually shifted to the view that it does not strictly matter for any tangible reason whether you get the full size you request or whether the volume you get is actually short -- even by as much as 100 MiB in most cases. Of course not everyone will be pleased by such a convention, but in this role we have a mandate to make it easy to use for the majority of cases. We could also introduce a switch to enable/disable size fuzzing if we want to try to please everyone.

vojtechtrefny:
Closed via #199

PR #189, opened 2021-01-21
Add new test case for bz1894647

gitPo1son:
Playbook will fail when lv size is equal to vg size

Signed-off-by: Changhui Zhong <czhong@redhat.com>

PR #181, opened 2020-12-08
LVMVDO support

japokorn:
- added support for LVM VDO
- new accepted pool options:
    - vdo_compression
    - vdo_deduplication
    - vdo_size
- added tests for LVM VDO

PR #199, opened 2021-02-11
Trim volume size as needed to fit in pool free space

dwlehman:
I've included the test from #189 since it is intended to verify this change.

I am open to a discussion of whether we should expose `MAX_TRIM_PERCENT` in the role defaults. I'm also open to discussion of what the best default value is. I would definitely be open to reducing it to 2-4% as that's more in line with my intended application of this.

### DIFF
--- a/tests/tests_lvm_auto_size_cap.yml
+++ b/tests/tests_lvm_auto_size_cap.yml
@@ -1,0 +1,91 @@
+---
+- hosts: all
+  become: true
+
+  tasks:
+    - include_role:
+        name: linux-system-roles.storage
+
+    - include_tasks: get_unused_disk.yml
+      vars:
+        min_size: 10g
+        max_return: 1
+
+    - command: lsblk -b -l --noheadings -o NAME,SIZE
+      register: storage_test_lsblk
+      changed_when: false
+
+    - set_fact:
+        test_disk_size: "{{ storage_test_lsblk.stdout_lines|map('regex_search', '^' + unused_disks[0] + '\\s+\\d+$')|select('string')|first|regex_replace('^\\w+\\s+', '') }}"
+
+    - package:
+        name: bc
+        state: installed
+
+    - command:
+        cmd: bc
+        stdin: "{{ test_disk_size }} *2"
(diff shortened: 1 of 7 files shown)


dwlehman:
This should resolve #13

dwlehman:
[citest bad]

richm:
[citest bad]
staging now has the fix so that tests from other PRs won't be used

richm:
[citest bad]

nhosoi:
[citest bad]
(let me rerun the CI to test https://issues.redhat.com/browse/RHELPLAN-67929.)

nhosoi:
[citest pending]

nhosoi:
[citest]

nhosoi:
> [citest bad]
> staging now has the fix so that tests from other PRs won't be used

We could verify that there is no more extra files from previous PRs, e.g., with  `rhel-x/ansible-2.9 (staging) — linux-system-roles-staging-141-jbrp5@2021-02-19 07:06:56.290667`. It ran after the new test-harness was deployed as well as after a pr with extra test (#181).

tests/tests_lvm_auto_size_cap.yml
```diff
@@ -0,0 +1,89 @@
...
+    - include_role:
+        name: linux-system-roles.storage
+
+    - include_tasks: get_unused_disk.yml
+      vars:
+        min_size: 10g
+        max_return: 1
+
+    - command: lsblk -b -l --noheadings -o NAME,SIZE
+      register: storage_test_lsblk
```
japokorn, line 15: Will adding `changed_when: false` resolve this?

japokorn (commented):
One comment, otherwise it looks good to me.

tests/tests_lvm_auto_size_cap.yml
richm: yes - please add `changed_when: false`

dwlehman:
Latest revision added `changed_when: false` to two plays(?) in the tests and a little blurb to the README about the size trimming.

PR #200, opened 2021-02-12
remove include_vars, use public: true to export role vars, defaults

richm:
The tests_deps.yml test was using include_vars assuming the symlink
roles/linux-system-roles.storage was present.  This is deprecated.
The test was doing this in order to get the defaults and vars from
the role.  The better way to do this is to use `include_role` with
the `public: true` parameter so that the role will expose its private
variables to the play.


nhosoi (approved):
lgtm

richm:
[citest bad]

PR #210, opened 2021-04-07
Remove python-26 environment from tox testing

richm:
This role does not currently need tox testing on python 2.6, which
is primarily for modules that run on EL6 managed hosts.  Testing on
python 2.6 is problematic due to the age and supportability, which
causes false positive tests, cluttering up the status and adding
noise.  Therefore, we should get rid of the test.

If the role in the future needs to test with python 2.6, see
https://github.com/linux-system-roles/lsr-gh-action-py26


richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

PR #212, opened 2021-04-12
This is just to test CI can handle it [WIP]

japokorn:
This branch is not to be merged.
It will instruct the CI to run the tests with VM with nine disks just to
make sure it can handle it.


japokorn:
All the tests have passed. CI can indeed handle a VM with nine disks.

PR #215, opened 2021-07-22
skip vdo test if kernel module is not loadable

richm:
The vdo kernel module is an "out of tree" module which means that it
may not be available if the vdo module version does not match the
current kernel version.  In that case, skip the vdo test so that we
don't get a false positive test result.


richm:
@vojtechtrefny fyi

dwlehman (approved):
I think the four added files in tests/ should be dropped, but otherwise it looks good to me.

japokorn (approved):
Looks good to me.

Generated vdo test files for nvme and iscsi are currently missing in the main branch and I believe they should be added there.

dwlehman:
> Generated vdo test files for nvme and iscsi are currently missing in the main branch and I believe they should be added there.

Ah, ok. Makes sense to me.

richm:
[citest bad] - some weird flakes on rhel7 and rhel9

richm:
@dwlehman @japokorn yay - rhel-x tests passed - ready to merge?

PR #217, opened 2021-07-28
omit unnecessary conditional - deadcode reported by static scanner

richm:
The list comprehension does not need the conditional evaluation since
the condition is already checked by the `if` clause.  This code
triggers a warning by a static code checker.

Also adds the scsi and nvme tests for tests_lvm_percent_size
which was added by an earlier commit.

https://bugzilla.redhat.com/show_bug.cgi?id=1985572


spetrosi (approved):
lgtm

PR #220, opened 2021-08-06
use volume1_size; check for expected error

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1990793
The test was using `volume_size`, which is undefined, instead
of `volume1_size`.  Furthermore, the error was being masked by
the `rescue` block because the rescue was not checking for the
correct error message.  The fix is to use `volume1_size` AND check
for the correct error message.

I'll note that there are several tests that do not look for the
correct error message, and we may run into this problem with other
tests.

### DIFF
--- a/tests/tests_lvm_errors.yml
+++ b/tests/tests_lvm_errors.yml
@@ -103,7 +103,7 @@
                 disks: "{{ unused_disks[0] }}"
                 volumes:
                   - name: test1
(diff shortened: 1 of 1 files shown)


richm:
failures are known or are network problems - I see the "Verify the output" in the tests_lvm_errors test - merging

PR #223, opened 2021-08-20
Nested module params checking

japokorn:
Older versions of ansible do not validate nested modules parameters for
types etc. Ansible 2.11 and newer implement this functionality. This
commit adds type definitions for nested variables.

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -144,6 +144,7 @@ except ImportError:
         LIB_IMP_ERR = traceback.format_exc()
 
 from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.storage_lsr.argument_validator import validate_parameters
 
 if BLIVET_PACKAGE:
     blivet_flags.debug = True
@@ -411,7 +412,7 @@ class BlivetVolume(BlivetBase):
     def _apply_defaults(self):
         global volume_defaults
         for name, default in volume_defaults.items():
-            if name in self._volume:
+            if name in self._volume and self._volume[name] not in [None, list(), dict()]:
                 continue
 
             default = None if default in ('none', 'None', 'null') else default
@@ -421,7 +422,8 @@ class BlivetVolume(BlivetBase):
                 if not self._update_from_device(name):
                     self._volume[name] = default
             else:
-                self._volume.setdefault(name, default)
+                if self._volume.get(name) is None:
+                    self._volume[name] = default
 
     def _get_format(self):
         """ Return a blivet.formats.DeviceFormat instance for this volume. """
@@ -862,7 +864,10 @@ _BLIVET_VOLUME_TYPES = {
 def _get_blivet_volume(blivet_obj, volume, bpool=None):
     """ Return a BlivetVolume instance appropriate for the volume dict. """
     global volume_defaults
-    volume_type = volume.get('type', bpool._pool['type'] if bpool else volume_defaults['type'])
+
+    volume_type = volume.get('type', None)
+    if volume_type is None:
+        volume_type = bpool._pool['type'] if bpool else volume_defaults['type']
     if volume_type not in _BLIVET_VOLUME_TYPES:
         raise BlivetAnsibleError("Volume '%s' has unknown type '%s'" % (volume['name'], volume_type))
 
@@ -1020,7 +1025,7 @@ class BlivetPool(BlivetBase):
     def _apply_defaults(self):
         global pool_defaults
         for name, default in pool_defaults.items():
-            if name in self._pool:
+            if name in self._pool and self._pool[name] not in [None, list(), dict()]:
                 continue
 
             default = None if default in ('none', 'None', 'null') else default
@@ -1029,7 +1034,8 @@ class BlivetPool(BlivetBase):
                 if not self._update_from_device(name):
                     self._pool[name] = default
             else:
-                self._pool.setdefault(name, default)
+                if self._pool.get(name) is None:
+                    self._pool[name] = default
 
     def _create_members(self):
         """ Schedule actions as needed to ensure pool member devices exist. """
@@ -1187,7 +1193,7 @@ _BLIVET_POOL_TYPES = {
 
 def _get_blivet_pool(blivet_obj, pool):
     """ Return an appropriate BlivetPool instance for the pool dict. """
-    if 'type' not in pool:
+    if 'type' not in pool or pool['type'] is None:
         global pool_defaults
         pool['type'] = pool_defaults['type']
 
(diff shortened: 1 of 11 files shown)


richm:
> Older versions of ansible do not validate nested modules parameters for
> types etc.

Unfortunately we have to support those older versions of ansible for quite some time . . .

> Ansible 2.11 and newer implement this functionality. This
> commit adds type definitions for nested variables.

richm:
[citest bad]

richm:
[citest bad]

library/blivet.py
```diff
@@ -411,7 +412,7 @@ def _update_from_device(self, param_name):
     def _apply_defaults(self):
         global volume_defaults
         for name, default in volume_defaults.items():
-            if name in self._volume:
+            if name in self._volume and self._volume[name] not in [None, list(), dict()]:
```
richm, line 415: what does it mean for `self._volume[name]` to be `None`?  That is - that appears to be a valid configuration?  Does it mean "use the defaults"?
I think you want to use `isinstance(self._volume[name], list, dict, None)` instead of what you have here

richm: If `self._volume[name]` is not None, a dict, or a list, should that be an error for the user?

library/blivet.py
```diff
@@ -1401,14 +1407,65 @@ def activate_swaps(b, pools, volumes):
...
+                                  size=dict(type='str'),
+                                  state=dict(type='str', default='present', choices=['present', 'absent']),
+                                  type=dict(type='str'))),
         packages_only=dict(type='bool', required=False, default=False),
         disklabel_type=dict(type='str', required=False, default=None),
         safe_mode=dict(type='bool', required=False, default=True),
         pool_defaults=dict(type='dict', required=False),
         volume_defaults=dict(type='dict', required=False),
-        use_partitions=dict(type='bool', required=False, default=True),
+        use_partitions=dict(type='bool', required=False),
```
richm, line 1468: This means the default value for `use_partitions` is `None` - and `if use_partitions:` will return False if `use_partitions` is either `False` or `None` - do you need to distinguish the case where the user does not set `use_partitions` vs. the user explicitly sets `use_partitions: False`?

module_utils/storage_lsr/argument_validator.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class ArgValidator(object):
+
+    @classmethod
+    def _validate_list(cls, value):
+        if isinstance(value, list):
+            return value
+
+        if isinstance(value, str):
```
richm, line 13: It is a bit complicated to test if an object is a string and make it portable for both python2 and python3
see https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/utils.py#L20-L21
and
https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/utils.py#L280

module_utils/storage_lsr/argument_validator.py
```diff
@@ -0,0 +1,171 @@
...
+        raise TypeError("'%s' has to be a list" % value)
+
+    @classmethod
+    def _validate_str(cls, value):
+        return str(value)
+
+    @classmethod
+    def _validate_bool(cls, value):
+        BOOLEANS_TRUE = ['y', 'yes', 'on', '1', 'true', 't', 1, 1.0, True]
+        BOOLEANS_FALSE = ['n', 'no', 'off', '0', 'false', 'f', 0, 0.0, False]
```
richm, line 28: Do you really want or need to support all of these different values?  Note that Ansible/Yaml will convert values `value: off` to python `False` - and if someone explicitly specifies something like `value: "off"` meaning the string value "off" instead of the boolean value - I think that should (usually) be an error.
Also note that there is the function https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool which can handle most of these - and I don't think you need to worry about `float` `0.0` and `1.0`

module_utils/storage_lsr/argument_validator.py
```diff
@@ -0,0 +1,171 @@
...
+            return value.split(",")
+
+        if isinstance(value, int) or isinstance(value, float):
+            return [value]
+
+        raise TypeError("'%s' has to be a list" % value)
+
+    @classmethod
+    def _validate_str(cls, value):
+        return str(value)
```
richm, line 23: see other comment about strings

module_utils/storage_lsr/argument_validator.py
```diff
@@ -0,0 +1,171 @@
...
+            return False
+        if value is None:
+            return None
+        raise TypeError("'%s' has to be convertable to a bool" % value)
+
+    @classmethod
+    def _validate_int(cls, value):
+        if isinstance(value, int):
+            return value
+        if isinstance(value, str):
```
richm, line 42: see other comment about strings
also, can't you just do something like this?:
```

    def _validate_int(cls, value):
        try:
            return int(value)
        except ValueError:
            pass
        raise TypeError("'%s' has to be an integer" % value)
```

module_utils/storage_lsr/argument_validator.py
```diff
@@ -0,0 +1,171 @@
...
+                return int(value)
+            except ValueError:
+                pass
+        raise TypeError("'%s' has to be an integer" % value)
+
+    @classmethod
+    def _validate_float(cls, value):
+        if isinstance(value, float):
+            return value
+        if isinstance(value, str):
```
richm, line 53: see comment above

module_utils/storage_lsr/argument_validator.py
japokorn: In this check, no actual validation is done, because anything can be converted to string. This will just make sure that the no matter what input was the result is string type.
This should be doable by standard `str()`.

module_utils/storage_lsr/argument_validator.py
japokorn: Thank you. Fixed.

module_utils/storage_lsr/argument_validator.py
japokorn: Certainly much more elegant solution. Fixed.

module_utils/storage_lsr/argument_validator.py
japokorn: Fixed.

japokorn:
[citest bad]

japokorn:
[citest bad]

richm (approved):
lgtm - don't worry about the CI failures - enough of them passed to know this is good

richm:
@japokorn can you rebase this PR on top of the latest master branch?  I'd like to make sure this PR is ready for Ansible 2.12

spetrosi:
[citest]

module_utils/storage_lsr/argument_validator.py
```diff
@@ -0,0 +1,168 @@
+from __future__ import absolute_import, division, print_function
+
+import sys
+
+__metaclass__ = type
+
+# pylint: disable=undefined-variable
+STRING_TYPE = str if sys.version_info.major == 3 else basestring  # noqa:F821
```
richm, line 8: This is causing the ansible-test error:
```
Running sanity test "import" on Python 2.6
ERROR: Found 1 import issue(s) on python 2.6 which need to be resolved:
ERROR: plugins/module_utils/storage_lsr/argument_validator.py:8:0: traceback: AttributeError: 'tuple' object has no attribute 'major'
```
I guess `sys.version_info` is just a plain tuple on python 2.6

The solution is to add the following line to all of the .sanity-ignore*.txt files:
```
plugins/module_utils/storage_lsr/argument_validator.py import-2.6!skip
```

richm:
[citest bad]

one minor nit, otherwise, lgtm

japokorn:
> [citest bad]
> 
> one minor nit, otherwise, lgtm

Fixed.

japokorn:
[citest]

PR #227, opened 2021-09-14
replace json_query with selectattr and map

richm:
This is to prepare for ansible-core, which does not support `json_query`


spetrosi (approved):
lgtm

PR #228, opened 2021-09-17
Replace crypttab with lineinfile

spetrosi:
Crypttab is not available in ansible-core, hence need to replace it

### DIFF
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -145,7 +145,11 @@
 - name: manage /etc/crypttab to account for changes we just made
   lineinfile:
     path: /etc/crypttab
+    # The line option is used when `state: present`
     line: "{{ entry.name }} {{ entry.backing_device }} {{ entry.password }}"
+    # The regexp option is used when `state: absent` to remove entries
+    # regardless of password
+    regexp: ^{{ entry.name }} {{ entry.backing_device }}
     state: "{{ entry.state }}"
     create: true
     mode: "{{ __storage_crypttab.stat.mode | d('0600') }}"
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -137,11 +137,20 @@
 #
 # Manage /etc/crypttab
 #
-- name: Manage /etc/crypttab to account for changes we just made
+- name: retrieve facts for the /etc/crypttab file
+  stat:
+    path: /etc/crypttab
+  register: __storage_crypttab
+
+- name: manage /etc/crypttab to account for changes we just made
   lineinfile:
     path: /etc/crypttab
     line: "{{ entry.name }} {{ entry.backing_device }} {{ entry.password }}"
     state: "{{ entry.state }}"
+    create: true
+    mode: "{{ __storage_crypttab.stat.mode | d('0600') }}"
+    owner: "{{ __storage_crypttab.stat.owner | d('root') }}"
+    group: "{{ __storage_crypttab.stat.group | d('root') }}"
   loop: "{{ blivet_output.crypts }}"
   loop_control:
     loop_var: entry
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -138,10 +138,9 @@
 # Manage /etc/crypttab
 #
 - name: Manage /etc/crypttab to account for changes we just made
-  crypttab:
-    name: "{{ entry.name }}"
-    backing_device: "{{ entry.backing_device }}"
-    password: "{{ entry.password }}"
+  lineinfile:
+    path: /etc/crypttab
+    line: "{{ entry.name }} {{ entry.backing_device }} {{ entry.password }}"
     state: "{{ entry.state }}"
   loop: "{{ blivet_output.crypts }}"
   loop_control:


tasks/main-blivet.yml
```diff
@@ -138,10 +138,9 @@
 # Manage /etc/crypttab
 #
 - name: Manage /etc/crypttab to account for changes we just made
-  crypttab:
-    name: "{{ entry.name }}"
-    backing_device: "{{ entry.backing_device }}"
-    password: "{{ entry.password }}"
+  lineinfile:
```
spetrosi, line 141: Hi @dwlehman, do you know if we can use the `cryptsetup` command for this? I didn't find a way to set password via it, and the `crypttab` [module](https://gitlab.ics.muni.cz/242618/ansible/-/blob/6a10514e086b839604d819cca8c2d19dd0362923/lib/ansible/modules/system/crypttab.py) simply handles the config file, so I ended up using `lineinfile`, but maybe you can help.
Thank you

spetrosi: Are the changes I made good to go then?

richm: > Are the changes I made good to go then?

I think so - let's see if we can get some better test results.

[citest]

richm:
[citest bad]

richm:
[citest pending]

richm:
[citest bad]

richm:
@spetrosi looks like the tests_luks.yml test is failing, and it looks like it has something to do with crypttab:
```
TASK [linux-system-roles.storage : Mask the systemd cryptsetup services] *******
task path: /tmp/tmpq1v69liz/tasks/main-blivet.yml:50

TASK [linux-system-roles.storage : manage the pools and volumes to match the specified state] ***
task path: /tmp/tmpq1v69liz/tasks/main-blivet.yml:56
fatal: [/cache/rhel-8-y.qcow2]: FAILED! => {"actions": [], "changed": false, "crypts": [], "leaves": [], "mounts": [], "msg": "encrypted volume 'foo' missing key/password", "packages": [], "pools": [], "volumes": []}
```
This is rhel 8.5 and rhel 9

tasks/main-blivet.yml
```diff
@@ -138,10 +138,9 @@
...
+    path: /etc/crypttab
+    line: "{{ entry.name }} {{ entry.backing_device }} {{ entry.password }}"
     state: "{{ entry.state }}"
```
richm, line 144: need to add `  create: true` so that the file will be created if it doesn't exist

tasks/main-blivet.yml
```diff
@@ -137,12 +137,18 @@
...
+    path: /etc/crypttab
+  register: __storage_crypttab
+
+- name: manage /etc/crypttab to account for changes we just made
+  lineinfile:
+    path: /etc/crypttab
+    line: "{{ entry.name }} {{ entry.backing_device }} {{ entry.password }}"
     state: "{{ entry.state }}"
+    create: true
+    mode: "{{ __storage_crypttab.stat.mode | d('0600') }}"
```
richm, line 151: and 
```
owner: "{{ __storage_crypttab.stat.owner | d('root') }}"
group: "{{ __storage_crypttab.stat.group | d('root') }}"
```

richm (approved):
lgtm

richm:
[citest]

richm:
There is definitely a bug with the new implementation.
I used a fedora 34 control node and a rhel-8-y managed host
Running the tests_luks.yml test on `master` branch works fine, but it breaks with this fix.  I'm not sure what the crypttab module is doing that makes it work.

spetrosi:
> There is definitely a bug with the new implementation. I used a fedora 34 control node and a rhel-8-y managed host Running the tests_luks.yml test on `master` branch works fine, but it breaks with this fix. I'm not sure what the crypttab module is doing that makes it work.

The issue was that the crypttab module allowed removing entries from /etc/crypttab without providing a password. I made lineinfile ignore the password when `state: absent`

richm:
[citest pending]

richm:
[citest bad]

nhosoi (approved):
lgtm

tasks/main-blivet.yml
```diff
@@ -145,10 +145,14 @@
...
     path: /etc/crypttab
+    # The line option is used when `state: present`
     line: "{{ entry.name }} {{ entry.backing_device }} {{ entry.password }}"
+    # The regexp option is used when `state: absent` to remove entries
+    # regardless of password
+    regexp: ^{{ entry.name }} {{ entry.backing_device }}
     state: "{{ entry.state }}"
     create: true
-    mode: "{{ __storage_crypttab.stat.mode | d('0600') }}"
+    mode: "{{ __storage_crypttab.stat.mode | d('0622') }}"
```
richm, line 155: this is an unusual mode - `-rw--w--w-` - the file has to be writable by all but not readable?

spetrosi: Sorry my bad, it should be `-rw-r--r--`, changed to 0644 now.

richm: does the file have to be world-readable?  It has a password in it?

spetrosi: On master, the role does `0600`. I mistakenly used `0644` because that's what lineinfile does by default. The file indeed stores an unencrypted password in it, so `0600` is the way to go. Sorry for the confusion and thank you for the thorough review!

spetrosi:
[citest]

richm:
[citest bad]

PR #229, opened 2021-09-24
Add support for Rocky Linux 8

darxriggs:


darxriggs:
It looks like that the same test failure is present for all tests. It is unrelated to this change (also present in other PRs) and should be fixed separately.

```
TASK [Exit playbook when there's not enough unused disks in the system] ********
fatal: [sut]: FAILED! => {
    "changed": false
}

MSG:

Unable to find enough unused disks. Exiting playbook.
	to retry, use: --limit @/tmp/tests_volume_relabel.retry
```

richm:
> It looks like that the same test failure is present for all tests. It is unrelated to this change (also present in other PRs) and should be fixed separately.
> 
> ```
> TASK [Exit playbook when there's not enough unused disks in the system] ********
> fatal: [sut]: FAILED! => {
>     "changed": false
> }
> 
> MSG:
> 
> Unable to find enough unused disks. Exiting playbook.
> 	to retry, use: --limit @/tmp/tests_volume_relabel.retry
> ```

Right.  Just ignore that for now.  We're working on fixing the tests to skip that.

spetrosi:
[citest]

darxriggs:
An alternative could also be to add `vars/RedHat.yml` as there is also the `"{{ ansible_facts['os_family'] }}.yml"` lookup (see  [tasks/main.yml](https://github.com/linux-system-roles/storage/blob/1.6.1/tasks/main.yml#L10)). This would support even more OS distributions.

richm:
> An alternative could also be to add `vars/RedHat.yml` (and consolidate the `CentOS.8.yml`, `RedHat.8.yml` and `RedHat.9.yml` files) as there is also the `"{{ ansible_facts['os_family'] }}.yml"` lookup (see [tasks/main.yml](https://github.com/linux-system-roles/storage/blob/1.6.1/tasks/main.yml#L10)). This would support even more OS distributions.

Does `ansible_facts['os_family']` on a Rocky Linux system return "RedHat"?  If so, do you know in which version of Ansible this is supported (if it depends on the version of Ansible - not sure how this fact works).

darxriggs:
Rocky Linux is supported since Ansible 2.11. `ansible_os_family` is set to `RedHat` in this case.
The change in the mapping from distribution to OS family can be seen in https://github.com/ansible/ansible/pull/74530/files.

Unfortunately due to the Ansible release cycle this will not be backported, see https://github.com/ansible/ansible/issues/74565.

But Ansible 2.9.23-1 in EPEL contains a change that adds the Rocky Linux support, see https://koji.fedoraproject.org/koji/buildinfo?buildID=1775015.

richm:
> Rocky Linux is supported since Ansible 2.11. `ansible_os_family` is set to `RedHat` in this case. The change in the mapping from distribution to OS family can be seen in https://github.com/ansible/ansible/pull/74530/files.

OK.  As soon as we can drop support for Ansible 2.9, we can switch here (and the other roles) to use `ansible_os_family`, but for now, let's keep this like you have it.

> 
> Unfortunately due to the Ansible release cycle this will not be backported, see [ansible/ansible#74565](https://github.com/ansible/ansible/issues/74565).
> 
> But Ansible 2.9.23-1 in EPEL contains a change that adds the Rocky Linux support, see https://koji.fedoraproject.org/koji/buildinfo?buildID=1775015.

richm:
[citest commit:e572ab58bee1e0490aa377f49f078464130e9c89]

richm:
[citest pending]

darxriggs:
I double checked with https://galaxy.ansible.com/docs/contributing/creating_role.html. Under "platforms" it references the Ansible Galaxy API. So I searched for Rocky via https://galaxy.ansible.com/api/v1/platforms/?name__icontains=Rocky with 0 results. There is also no "CentOS" platform. Therefore having `EL` only should be fine.

PR #230, opened 2021-09-25
Improve wording

darxriggs:


README-devel.md
```diff
@@ -43,7 +42,7 @@ There are unit tests for the embedded modules in `tests/unit/` and integration
 tests for the role in `tests/`.
 
 ## Running Tests
-All of the integration tests are designed to use extra/unused disks in a vm.
+All the integration tests are designed to use extra/unused disks in a vm.
```
richm, line 45: I prefer `All of`

README.md
```diff
@@ -50,7 +50,7 @@ This is a list of volumes that belong to the current pool. It follows the
 same pattern as the `storage_volumes` variable, explained below.
 
 ##### `encryption`
-This specifies whether or not the pool will be encrypted using LUKS.
+This specifies whether the pool will be encrypted using LUKS.
```
richm, line 53: I would prefer "This specifies that the pool . . ." - I prefer indicative rather than conditional or subjunctive

README.md
```diff
@@ -138,7 +138,7 @@ When type is `raid` specifies RAID chunk size as a string, e.g.: '512 KiB'.
 Chunk size has to be multiple of 4 KiB.
 
 ##### `encryption`
-This specifies whether or not the volume will be encrypted using LUKS.
+This specifies whether the volume will be encrypted using LUKS.
```
richm, line 141: I would prefer "This specifies that the pool . . ." - I prefer indicative rather than conditional or subjunctive

README.md
```diff
@@ -162,7 +162,7 @@ This integer specifies the LUKS key size (in bits).
 This integer specifies the LUKS version to use.
 
 ##### `deduplication`
-This specifies whether or not the Virtual Data Optimizer (VDO) will be used.
+This specifies whether the Virtual Data Optimizer (VDO) will be used.
```
richm, line 165: I would prefer "This specifies that the pool . . ." - I prefer indicative rather than conditional or subjunctive

README.md
```diff
@@ -171,7 +171,7 @@ Limit one VDO `storage_volume` per `storage_pool`.
 Underlying volume has to be at least 9 GB (bare minimum is around 5 GiB).
 
 ##### `compression`
-This specifies whether or not the Virtual Data Optimizer (VDO) will be used.
+This specifies whether the Virtual Data Optimizer (VDO) will be used.
```
richm, line 174: I would prefer "This specifies that the pool . . ." - I prefer indicative rather than conditional or subjunctive

README.md
darxriggs: Does "that ... will be encrypted" really cover the `false` value?

richm: hmm - probably not - so I guess it is ok as you have it

README.md
richm: I guess the same https://github.com/linux-system-roles/storage/pull/230#discussion_r716107461 applies here - so whether is more appropriate - but I'm not sure what happens if you specify `compression: false` on an existing volume/pool that is using VDO compression . . .

README.md
richm: I guess the same https://github.com/linux-system-roles/storage/pull/230#discussion_r716107461 applies here - so whether is more appropriate - but I'm not sure what happens if you specify `deduplication: false` on an existing volume/pool that is using VDO deduplication . . .

README-devel.md
darxriggs: I just reverted this change.

richm (approved):

PR #237, opened 2021-11-11
workaround for udev issue on some platforms

richm:
On some platforms, udev reports that the device is not in use when
it actually is being used.  This cause the idempotency checking part
of the test to fail.  The solution is to use `udevadm trigger` to
force udev to rescan and pick up any pending changes.

### DIFF
--- a/tests/tests_change_fs.yml
+++ b/tests/tests_change_fs.yml
@@ -23,6 +23,10 @@
         min_size: "{{ volume_size }}"
         max_return: 1
 
+    - name: Workaround for udev issue on some platforms
+      command: udevadm trigger --subsystem-match=block
+      changed_when: false
+
     - name: Create a LVM logical volume with default fs_type
       include_role:
         name: linux-system-roles.storage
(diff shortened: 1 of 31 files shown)


dwlehman (approved):
lgtm

richm:
[citest]

richm:
[citest]

richm:
[citest bad]

richm:
[citest bad]

richm:
[citest bad]

richm:
@dwlehman kinda ugly, but seems to make basic smoke test happy on rhel9

richm:
[citest]

richm:
[citest bad]

richm:
@dwlehman I had to add a udevadm to almost every test in basic smoke test - this seems excessive - maybe there is a better way?

richm:
[citest bad]

richm:
@dwlehman I would like to close this PR and use this one instead - https://github.com/linux-system-roles/storage/pull/243 - if the issue is that the changes that blivet makes aren't immediately recognized by udev, then it seems the right way to handle this is to use udevadm to force udev to recognize those changes immediately after blivet changes something.

PR #241, opened 2021-11-23
Added support for LVM RAID volumes

japokorn:
- raid_level can be used on pool volume
- raid_disks can specify disks to be used
- test added
- refactored BlivetLVMVolume._create
- some changes to README-devel (more to come)

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -665,73 +665,83 @@ class BlivetLVMVolume(BlivetVolume):
             size = self._blivet_pool._device.align(size, roundup=True)
         return size
 
-    def _create(self):
-        if self._device:
-            return
+    def _trim_size(self, size):
+        parent_device = self._blivet_pool._device
 
-        parent = self._blivet_pool._device
-        if parent is None:
-            raise BlivetAnsibleError("failed to find pool '%s' for volume '%s'" % (self._blivet_pool['name'], self._volume['name']))
-
-        size = self._get_size()
-
-        fmt = self._get_format()
-        trim_percent = (1.0 - float(parent.free_space / size)) * 100
+        trim_percent = (1.0 - float(parent_device.free_space / size)) * 100
         log.debug("size: %s ; %s", size, trim_percent)
 
-        create_vdo = self._volume['deduplication'] or self._volume['compression']
-
-        if create_vdo:
-            pool_size = parent.free_space if size > parent.free_space else size
-            if self._volume['vdo_pool_size']:
-                pool_size = Size(self._volume['vdo_pool_size'])
-                if pool_size > parent.free_space:
-                    if trim_percent > MAX_TRIM_PERCENT:
-                        raise BlivetAnsibleError("specified 'vdo_pool_size' for volume '%s' "
-                                                 "exceeds available space in pool '%s' (%s)" % (pool_size,
-                                                                                                parent.name,
-                                                                                                parent.free_space))
-                    else:
-                        log.info("adjusting %s size from %s to %s to fit in %s free space",
-                                 self._volume['name'],
-                                 size,
-                                 parent.free_space,
-                                 parent.name)
-                        pool_size = parent.free_space
-
-            try:
-                vdopool = self._blivet.new_lv(name='vdopool', vdo_pool=True,
-                                              parents=[parent], compression=self._volume['compression'],
-                                              deduplication=self._volume['deduplication'],
-                                              size=pool_size)
-            except Exception as e:
-                raise BlivetAnsibleError("failed to set up VDO pool '%s': %s" % (self._volume['name'], str(e)))
-
-            self._blivet.create_device(vdopool)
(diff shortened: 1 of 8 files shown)


tests/tests_create_lvmraid_then_remove.yml
```diff
@@ -0,0 +1,74 @@
+---
```
japokorn, line 1: For the sake of test speed I should probably make this test a part of one that already exists.

japokorn:
[citest bad]

tests/tests_create_lvmraid_then_remove.yml
japokorn: Fixed. Test file removed, contents appended to `tests_create_raid_pool_then_remove.yml`.

richm:
lgtm - @dwlehman ?
also - if you rebase on top of the latest `master` branch you can get rid of the tox errors

japokorn:
rebased

library/blivet.py
```diff
@@ -665,73 +665,101 @@ def _get_size(self):
...
+        parent_device = self._blivet_pool._device
+
+        if self._volume['raid_disks']:
+            pvs = list()
+
+            for path in self._volume['raid_disks']:
+                for pv in parent_device.pvs:
+                    disk_path = pv.path if pv.is_disk else pv.disk.path
+                    # We need to ensure we match '/dev/sda' with 'sda' but not with 'da'
+                    if ('/' + disk_path).endswith('/' + path):
```
vojtechtrefny, line 726: You can use `os.path.basename` to get name of the device from path, but I would suggest using `resolve_device` to get blivet devices for the `raid_disks` and then compare these to ancestors of the PVs.

library/blivet.py
```diff
@@ -665,73 +665,101 @@ def _get_size(self):
...
 
-        size = self._get_size()
+        use_vdo = self._volume['deduplication'] or self._volume['compression']
+        use_lvmraid = self._volume['raid_level']
 
-        fmt = self._get_format()
-        trim_percent = (1.0 - float(parent.free_space / size)) * 100
-        log.debug("size: %s ; %s", size, trim_percent)
+        # VDO size is technically unlimited, so no size checks needed when it is used
+        size = self._get_size() if use_vdo else self._trim_size(self._get_size())
```
vojtechtrefny, line 745: Please use "full" `if-else` here, we don't need to save space and using function call as an function argument is very hard to read and debug, it's not saving worth the two or three extra lines here.

library/blivet.py
```diff
@@ -665,73 +665,101 @@ def _get_size(self):
...
+            else:
+                log.info("adjusting %s size from %s to %s to fit in %s free space",
+                         self._volume['name'],
+                         size,
+                         parent_device.free_space,
+                         parent_device.name)
+                return parent_device.free_space
+        return size
+
+    def _create_vdo(self):
```
vojtechtrefny, line 688: These two functions (`_create_vdo` and `_create_lvmraid`) are really confusing -- these should either create everything or return the extra arguments for `new_lv`, now `_create_vdo` does a little bit of both (creates the pool and returns argument for the vdolv) and `_create_lvmraid` only returns extra arguments for `new_lv`. I think actually creating the LVs in these two functions would be better.

vojtechtrefny (changes requested):
I also think this should be split into multiple commits -- the refactoring of `BlivetLVMVolume._create` should be a separate commit, it shouldn't be mixed with the LVM RAID support implementation. It would be easier to review this way and it also make sense for future git blames.

library/blivet.py
japokorn: Renamed methods to less confusing names and added explanatory comments.

The goal is to allow simple adding of similar functions in the future - each of them producing its own parameters for `new_lv` call in `_create`. This also allows to handle all these potential functions as NOT mutual exclusive. (Ideally the mutual exclusivity would be taken care of much earlier - in the parameter checking.)
Hence the similar names of these functions despite their inner workings. Originally I named them `get_[vdo/lvmraid]_parameters` (which is these functions' primary purpose) but VDO also requires to have additional LV `vdopool` created, so...

library/blivet.py
japokorn: Fixed.

library/blivet.py
japokorn: Fixed.

japokorn:
> I also think this should be split into multiple commits -- the refactoring of `BlivetLVMVolume._create` should be a separate commit, it shouldn't be mixed with the LVM RAID support implementation. It would be easier to review this way and it also make sense for future git blames.

Fixed. Splitted PR into two commits.

richm:
@vojtechtrefny ok to merge?

vojtechtrefny (approved):
Looks good now, thank you.

japokorn:
Merged.

PR #242, opened 2021-11-29
add tags to allow skipping lvm tests

richm:
lvm is currently broken on el9.  We need to skip these tests until
lvm is fixed.  See https://bugzilla.redhat.com/show_bug.cgi?id=1967212

### DIFF
--- a/tests/tests_change_fs.yml
+++ b/tests/tests_change_fs.yml
@@ -6,6 +6,8 @@
     mount_location: '/opt/test1'
     volume_size: '5g'
     fs_after: "{{ (ansible_distribution == 'RedHat' and ansible_distribution_major_version == '6') | ternary('ext4', 'xfs') }}"
+  tags:
+    - tests::lvm
 
   tasks:
     - include_role:
--- a/tests/tests_change_fs_use_partitions.yml
+++ b/tests/tests_change_fs_use_partitions.yml
@@ -7,6 +7,8 @@
     mount_location: '/opt/test1'
     volume_size: '5g'
     fs_type_after: "{{ 'ext3' if (ansible_distribution == 'RedHat' and ansible_distribution_major_version == '6') else 'ext4' }}"
+  tags:
+    - tests::lvm
 
   tasks:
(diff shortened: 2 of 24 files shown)


richm:
@vojtechtrefny

richm:
Note for reviewers - the default `type` for a volume is `lvm` - so if the test does not specify `type: something` in the `vars`, then it defaults to `lvm` - this is why some tests appear to be skipped without specifying `type: lvm` explicitly

nhosoi:
Do we need `tests::lvm` tag for https://github.com/richm/storage/blob/tag-lvm-tests/tests/tests_deps.yml, too?

richm:
> Do we need `tests::lvm` tag for https://github.com/richm/storage/blob/tag-lvm-tests/tests/tests_deps.yml, too?

I don't think so, because that test only uses `blivet` in `packages_only` mode, which doesn't actually modify the lvm configuration (storage team, please correct me if I'm wrong)

richm:
[citest bad]

nhosoi (approved):
Thanks for the clarification, @richm.

lgtm

richm:
. . . and now the rhel-x test is inexplicably passing :-(  Was blivet updated?  was lvm2 reverted to not use the devices file?????

richm:
so some systems have `lvm2-2.03.14-2` which fails:
```
%changelog
* Thu Nov 18 2021 Marian Csontos <mcsontos@redhat.com> - 2.03.14-2
- Change use_devices_file default back to 1.
```
and some systems (like the one used by this CI) have `lvm2-2.03.14-1` which passes

vojtechtrefny:
> . . . and now the rhel-x test is inexplicably passing :-( Was blivet updated? was lvm2 reverted to not use the devices file?????

Yes the device file feature is disabled with lvm2-2.03.14-1, this is a [bug](https://bugzilla.redhat.com/show_bug.cgi?id=2021703) and it is fixed by lvm2-2.03.14-2 where the feature is enabled again.

PR #256, opened 2022-02-21
Less verbosity by default

lucab85:
Set the verbosity level to 1 for all the debug operation


richm:
[citest commit:0d188318063ce75b90e58e0757169f08d3462b30]

vojtechtrefny (approved):
I think we can hide these by default, these are mostly for debugging and we are running the tests with `-vv` in the CI anyway.

richm (approved):
lgtm

PR #262, opened 2022-04-08
allow role to work with gather_facts: false

richm:
Allow the role to work if the playbook has set `gather_facts: false`.
The `gather_subset: min` includes all of the facts used by this
role.  Other roles may need to gather more facts.
The role gathers the facts where they are needed.  Other roles may
need to gather facts in a different place.

### DIFF
--- a/tasks/set_vars.yml
+++ b/tasks/set_vars.yml
@@ -1,4 +1,10 @@
 ---
+- name: Ensure ansible_facts used by role
+  setup:
+    gather_subset: min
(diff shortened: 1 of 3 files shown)


nhosoi (approved):
lgtm

A good candidate to add to the template and the best-practices?

richm:
> lgtm
> 
> A good candidate to add to the template and the best-practices?

Yes.

vojtechtrefny (approved):
Looks good to me.

PR #263, opened 2022-04-18
add xfsprogs for non-cloud-init systems

richm:
There should be an explicit dependency on the `xfsprogs` packages.
The role works in CI because the CI systems are using `cloud-init`
based VMs which pull in xfsprogs.  On non-cloud-init systems the
package is missing and the role fails.


vojtechtrefny (approved):

richm:
[citest bad]

PR #269, opened 2022-05-06
Thin pool support

japokorn:
Storage role support for thin provisioning.
Volumes under LVM pool now can use three new options:
 - 'thin' - type bool
          - setting to True puts the volume into a thin pool
 - 'thin_pool_name' - type str
          - specifies the name of the thin pool for this volume
          - selects thin pool with the same name if it exists
          - tries to create new thin pool if it does not
          - if omitted and no existing thin pool is present, the name is generated automatically
          - if omitted and exactly one thin pool is present, the volume is put into existing thin pool
          - if omitted and more than one thin pool is present, exception is raised
 - 'thin_pool_size' - type Size
          - specifeis size of the thin pool

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -1,5 +1,4 @@
 #!/usr/bin/python
-
 from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
@@ -120,6 +119,7 @@ try:
     from blivet3.callbacks import callbacks
     from blivet3 import devices
     from blivet3.deviceaction import ActionConfigureFormat
+    from blivet3.devicefactory import DEFAULT_THPOOL_RESERVE
     from blivet3.flags import flags as blivet_flags
     from blivet3.formats import get_format
     from blivet3.partitioning import do_partitioning
@@ -134,6 +134,7 @@ except ImportError:
         from blivet.callbacks import callbacks
         from blivet import devices
         from blivet.deviceaction import ActionConfigureFormat
+        from blivet.devicefactory import DEFAULT_THPOOL_RESERVE
         from blivet.flags import flags as blivet_flags
         from blivet.formats import get_format
         from blivet.partitioning import do_partitioning
@@ -676,17 +677,17 @@ class BlivetLVMVolume(BlivetVolume):
             size = self._blivet_pool._device.align(size, roundup=True)
         return size
 
-    def _trim_size(self, size):
-        parent_device = self._blivet_pool._device
+    def _trim_size(self, size, parent_device):
 
         trim_percent = (1.0 - float(parent_device.free_space / size)) * 100
         log.debug("size: %s ; %s", size, trim_percent)
 
         if size > parent_device.free_space:
             if trim_percent > MAX_TRIM_PERCENT:
-                raise BlivetAnsibleError("specified size for volume '%s' exceeds available space in pool '%s' (%s)" % (size,
-                                                                                                                       parent_device.name,
-                                                                                                                       parent_device.free_space))
+                raise BlivetAnsibleError("specified size for volume '%s' '%s' exceeds available space in pool '%s' (%s)" % (self._volume['name'],
+                                                                                                                            size,
+                                                                                                                            parent_device.name,
+                                                                                                                            parent_device.free_space))
             else:
                 log.info("adjusting %s size from %s to %s to fit in %s free space",
                          self._volume['name'],
@@ -705,7 +706,7 @@ class BlivetLVMVolume(BlivetVolume):
 
         if self._volume['vdo_pool_size']:
             try:
-                pool_size = self._trim_size(Size(self._volume['vdo_pool_size']))
+                pool_size = self._trim_size(Size(self._volume['vdo_pool_size']), parent_device)
             except BlivetAnsibleError:
                 # Literally the same error just more specific description
                 raise BlivetAnsibleError("specified 'vdo_pool_size' for volume '%s' "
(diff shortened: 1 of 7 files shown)


library/blivet.py
```diff
@@ -791,29 +791,53 @@ def _get_params_lvmcache(self):
 
         return dict(cache_request=cache_request)
 
+    def _get_parent_thinpool_by_name(self, thin_pool_name):
+
+        if thin_pool_name is None:
+            if len(self._blivet_pool._thinpools) != 1:
+                raise BlivetAnsibleError("cannot determine thinpool for assigning volume '%s' " % self._volume.get('name'))
+            else:
+                return self._blivet_pool._thinpools[0]
```
richm, line 800: in the case where `len(self._blivet_pool._thinpools) == 1`, is it always true that `thin_pool_name == self._blivet_pool._thinpools[0].name`?

richm:
There's is some quite complex python code added . . .
For the blivet related code - @vojtechtrefny could you take a look?
For the validation code - would be really nice to have some sort of unit tests for this (and yes, I realize we have very little unit testing code as it is . . .)

library/blivet.py
japokorn: Yes. BlivetLVMVolume._get_parent_thinpool_by_name is called from BlivetLVMVolume._create under condition that thin pool creation is requested.
But before BlivetLVMVolume._create is called BlivetLVMPool._manage_thin_pools is executed which creates thin pools (and/or finds the existing ones).

japokorn:
@richm Makes sense. At this moment argument validation testing relies on the fact that our test suite uses wide variety of parameter combinations, hence it is at least 'somewhat' tested.
So my plan is to focus on extending 'regular' thin pool tests first, then move on to argument validation unit tests.

library/blivet.py
```diff
@@ -1265,18 +1289,98 @@ def _manage_encryption(self, members):
 
         return managed_members
 
+    def _manage_thin_pools(self, pool_device):
+
+        existing_thinlvs_obj = [d for d in pool_device.children if d.type == 'lvmthinpool']
```
vojtechtrefny, line 1294: You can use `is_thin_pool` instead of `d.type == 'lvmthinpool'`.

module_utils/storage_lsr/argument_validator.py
```diff
@@ -7,6 +7,26 @@
 # pylint: disable=undefined-variable
 STRING_TYPE = str if sys.version_info.major == 3 else basestring  # noqa:F821
 
+BOOLEANS_TRUE = ['y', 'yes', 'on', '1', 'true', 't', 1, 1.0, True]
```
vojtechtrefny, line 10: I think this should be a separate commit with some comment, this doesn't look related to the thin provisioning support.

japokorn: Fixed.

library/blivet.py
```diff
@@ -1265,18 +1289,98 @@ def _manage_encryption(self, members):
...
+                        reserved_space += Size(thin_size)
+            else:
+                # regular LV just take its size
+                vol_size = volume.get('size')
+                if vol_size is None:
+                    auto_size_dev_count += 1
+                else:
+                    reserved_space += Size(vol_size)
+
+        # Thin metadata will take 20% of VG space. At least 1GiB, at most 100GiB
```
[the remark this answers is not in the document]
japokorn: Thanks for clarification. Fixed.

library/blivet.py
```diff
@@ -791,29 +791,53 @@ def _get_params_lvmcache(self):
...
+
+        if thin_pool_name is None:
+            if len(self._blivet_pool._thinpools) != 1:
+                raise BlivetAnsibleError("cannot determine thinpool for assigning volume '%s' " % self._volume.get('name'))
+            else:
+                return self._blivet_pool._thinpools[0]
+
+        for thinp in self._blivet_pool._thinpools:
+            # lvm concatenates the name as <vg name>-<thin pool name>
+            if self._blivet_pool._device.name + '-' + thin_pool_name == thinp.name:
```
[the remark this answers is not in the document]
japokorn: Thanks. Fixed.

library/blivet.py
```diff
@@ -153,6 +152,7 @@
 
 
 MAX_TRIM_PERCENT = 2
+DEFAULT_THPOOL_RESERVE = dict(percent=20, min=Size("1 GiB"), max=Size("100 GiB"))
```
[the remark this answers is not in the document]
japokorn: It was a remnant from debugging. Thanks for noticing. Fixed.

library/blivet.py
```diff
@@ -1538,6 +1642,9 @@ def run_module():
                                                           raid_level=dict(type='str'),
                                                           size=dict(type='str'),
                                                           state=dict(type='str', default='present', choices=['present', 'absent']),
+                                                          thin_pool_name=dict(type='str'),
+                                                          thin_pool_size=dict(type='str'),
+                                                          thin=dict(type='bool', default=False),
```
richm, line 1647: You'll have to rebase this on top of my recent change

richm:
centos-7 tests are failing due to missing libraries:
```
** (process:2499): [1;33mWARNING[0m **: [34m13:18:30.150[0m: failed to load module btrfs: libbd_btrfs.so.2: cannot open shared object file: No such file or directory

** (process:2499): [1;33mWARNING[0m **: [34m13:18:30.150[0m: failed to load module loop: libbd_loop.so.2: cannot open shared object file: No such file or directory

** (process:2499): [1;33mWARNING[0m **: [34m13:18:30.151[0m: failed to load module mpath: libbd_mpath.so.2: cannot open shared object file: No such file or directory

** (process:2499): [1;33mWARNING[0m **: [34m13:18:30.152[0m: failed to load module nvdimm: libbd_nvdimm.so.2: cannot open shared object file: No such file or directory
```

library/blivet.py
```diff
@@ -1265,18 +1289,98 @@ def _manage_encryption(self, members):
...
+                        auto_size_dev_count += 1
+                    else:
+                        reserved_space += Size(thin_size)
+            else:
+                # regular LV just take its size
+                vol_size = volume.get('size')
+                if vol_size is None:
+                    auto_size_dev_count += 1
+                else:
+                    reserved_space += Size(vol_size)
```
richm, line 1331: This is causing the wrong error message to be thrown.  The tests are expecting an error from `_get_size()` like `"invalid percentage '2x%' size specified for volume '%s'"`.  Instead, the error message is more cryptic:
```
ValueError: Failed to recognize unit from the spec: 2x%
```
I think you are not supposed to use `Size()` directly unless you are sure the value is not a percentage.  I think you are supposed to use `Size()` directly unless you are sure the value is not a percentage.  I think you are supposed to use `self._get_size()`.  Unfortunately, I'm not sure which object to use for the `_get_size` method - the code is in `class BlivetLVMPool(BlivetPool)` but only `BlivetVolume` has the `_get_size` method.

Alternately you could change the test to look for the above error message, but IMO that isn't a good error message.

japokorn: Thank you. Fixed.

library/blivet.py
```diff
@@ -1265,18 +1289,114 @@ def _manage_encryption(self, members):
 
         return managed_members
 
+    def _manage_thin_pools(self, pool_device):
+
+        def get_vol_size(volume):
```
richm, line 1294: The parameter name should be `vol_size`, or `vol_size_str` if you think that would be more descriptive.

library/blivet.py
```diff
@@ -1265,18 +1289,114 @@ def _manage_encryption(self, members):
...
+            if isinstance(volume, str) and '%' in volume:
+                try:
+                    percentage = int(volume[:-1].strip())
+                except ValueError:
+                    raise BlivetAnsibleError("invalid percentage '%s' size specified for volume in pool '%s'" % (pool_device.name))
```
richm, line 1299: Error here - `TypeError: not enough arguments for format string` - you are missing the volume size str - so
```
                     raise BlivetAnsibleError("invalid percentage '%s' size specified for volume in pool '%s'" % (vol_size, pool_device.name))
```

richm:
any updates planned?

richm:
this lgtm - @vojtechtrefny ?

The test errors are due to CI issues

PR #270, opened 2022-05-09
add support for mount_options

richm:
When support for argument validation was added, that support did not
include the `mount_options` parameter.  This fix adds back that
parameter.  In addition, the volume module arguments are refactored
so that the common volume parameters such as `mount_options` can be
specified in one place.

This adds a test for the `mount_options` parameter, and adds
verification for that parameter.

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -105,6 +105,7 @@ volumes:
     elements: dict
 '''
 
+import copy
 import logging
 import os
 import traceback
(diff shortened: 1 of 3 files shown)


richm:
Why are the pool volume parameters (that is, the volume parameters specified under `pools`) different than the volume (top level) parameters?  Should they be different?  If so, does this PR have the correct separation of the pool volume from the volume parameters?

richm:
Note that most of the test failures are known, and none are related to this PR

japokorn:
@richm Yes, they should be different. There is actually a fair difference in internal workings of `volumes` and `pools.volumes`. We are trying to unify the interface for the user so the 'same' functionality is called in a same way. It works pretty well for basic parameters. More advanced features are usually specific to either pool or volume. (e.g. `volume.compression` is LVM feature and volume needs to be in 'lvm' type pool)
Maybe because of this we never actually changed it in `module_args`.

japokorn (approved):
Looks good to me.

PR #271, opened 2022-05-10
make striped an alias for raid0

richm:
blivet does not support a raid_level of `striped`, but it is the same
thing as `raid0`, so make `striped` an alias of `raid0`.

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -285,8 +285,11 @@ class BlivetBase(object):
             raise BlivetAnsibleError("chunk size must be multiple of 4 KiB")
 
         try:
+            raid_level = self._spec_dict["raid_level"]
+            if raid_level == "striped":
+                raid_level = "raid0"
             raid_array = self._blivet.new_mdarray(name=raid_name,
-                                                  level=self._spec_dict["raid_level"],
+                                                  level=raid_level,
                                                   member_devices=active_count,
                                                   total_devices=len(members),
                                                   parents=members,
@@ -740,7 +743,10 @@ class BlivetLVMVolume(BlivetVolume):
         else:
             pvs = parent_device.pvs
 
-        return dict(seg_type=self._volume['raid_level'], pvs=pvs)
+        seg_type = self._volume['raid_level']
+        if seg_type == "striped":
+            seg_type = "raid0"
+        return dict(seg_type=seg_type, pvs=pvs)
 
     def _detach_cache(self):
         """ Detach cache from the volume and remove the unused cache pool """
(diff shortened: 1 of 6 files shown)


library/blivet.py
```diff
@@ -801,6 +807,8 @@ def _create(self):
 
         use_vdo = self._volume['deduplication'] or self._volume['compression']
         use_lvmraid = self._volume['raid_level']
+        if use_lvmraid == "striped":
+            use_lvmraid = "raid0"
```
vojtechtrefny, line 811: This actually won't work because blivet doesn't support raid0 with LVM. This is a bug in blivet*, so this actually should be reversed for LVM RAID and we should replace `raid0` with `striped`.

*) To be more precise, it's a missing feature, support for `lvcreate --type raid0` was added later and we never added it to blivet.

vojtechtrefny:
I think this would be ok for a quick bug fix, but I think we need to revisit the RAID levels and test all the possible corner cases and decide what we should hide from the user.

LVM RAID still supports the legacy `mirror` and `striped` levels. We probably should remove these from the README, but the question is what should be done when user specifies these levels for LVM RAID:
-  When creating a new LV, should we create `mirror`/`striped` or `raid1`/`raid0` LV?
- If the LV already exists with `mirror`/`striped` and the playbook says `raid1`/`raid0` do we re-create it with the new level (if `safe_mode` is not enabled)?

The list of the accepted values for `raid_level` will also be more complicated in the future when/if we decide to add support to other technologies like for Stratis and Btrfs where the supported levels might be different.

vojtechtrefny:
To make my previous comment a little bit less confusing: RAID levels supported by LVM are: `striped`, `mirror`, `raid0`, `raid1` (plus the other `raidN` that are not part of this discussion). `striped` and `mirror` are old/obsolete and `raid0` and `raid1` should be preferred.

> How should we document this in the README? The current raid_level documentation implies that it only applies when `type: lvm`, and lists `striped` and `raid0` (and does not mention `mirror` at all).

I would probably remove `striped` from the documentation and mention only `raid0` and `raid1` (and other supported `raidN` options). (And deal with the fact that blivet doesn't support `raid0` internally.)

> When creating a new LV, should we create mirror/striped or raid1/raid0 LV?

By this I meant "What should we do when user says `mirror` or `striped`". Both LVM and blivet support these types so our options are:
- Create it even if our documentation doesn't mention these as supported. (`mirror` is not in README right now and I propose removing `striped` too.)
- Create `raid0`/`raid1` instead.
- Return an error saying these are not supported.

japokorn:
Note: blivet RAID0 support PR: https://github.com/storaged-project/blivet/pull/1047

richm:
ok - so what should we do for this issue?  The storage role needs to be able to handle the case where the user specifies `"raid0"` for the LVM raid type, but the version of blivet used does not support it.  Can we change the role so that when the user specifies `"raid0"`, we change it to `"striped"` before passing as an LVM raid type?

vojtechtrefny:
> Can we change the role so that when the user specifies `"raid0"`, we change it to `"striped"` before passing as an LVM raid type?

The problem is that `raid0` and `striped` are different segment types/levels in LVM. I proposed a different solution in #272, we don't need to wait for blivet to add the support, we can do that in our code.

PR #310, opened 2022-12-05
Thin pool test with large size volume fix

japokorn:
- fixed size calculation for large size thin pools in the test
- modified provision.fmf disk size to simulate larger disks in the tests


richm:
[citest]

japokorn:
[citest bad]

richm (approved):

PR #313, opened 2022-12-16
Use a threshold of 2 percent in volume size check

richm:
There seems to be an issue calculating the expected size and the
actual size of the volume.  On some systems, the difference is
greater than 1% but less than 2%.  We are working on a better, more
reliable method of calculating the expected and actual sizes.  In
the meantime, make the threshold 2%.


vojtechtrefny (approved):
I think this is a good workaround for now, thanks.

PR #331, opened 2023-01-31
restrict swap size to less than 128GB on EL7

richm:
On EL7 the swap size must be less than 128GB, so change the tests
to restrict the size.  This required introducing a `max_size` parameter
to find_unused_disk.

I also took this opportunity to do some ansible-lint 6.x cleanup
for the code I touched.  There is still a huge amount of tests/
code that is not ansible-lint 6.x clean.

### DIFF
--- a/library/find_unused_disk.py
+++ b/library/find_unused_disk.py
@@ -30,6 +30,11 @@ options:
         default: '0'
         type: str
 
+    max_size:
+        description: Specifies the maximum disk size to return an unused disk.
+        default: '0'
+        type: str
+
     with_interface:
         description: Specifies which disk interface will be accepted (scsi, virtio, nvme).
         default: null
@@ -157,6 +162,7 @@ def run_module():
     module_args = dict(
         max_return=dict(type='int', required=False, default=10),
         min_size=dict(type='str', required=False, default='0'),
+        max_size=dict(type='str', required=False, default='0'),
         with_interface=dict(type='str', required=False, default=None)
     )
 
@@ -170,6 +176,7 @@ def run_module():
         supports_check_mode=True
     )
 
+    max_size = Size(module.params['max_size'])
     for path, attrs in get_disks(module).items():
         if is_ignored(path):
             continue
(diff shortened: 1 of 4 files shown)


nhosoi:
lgtm

Out of curiosity... Instead of introducing `max_size` and passing it by setting `127g` to `find_unused_disk.py`, we could quietly assign `127g` to an internal variable for rhel-7, I'd think. Most likely, I haven't understood the advantage of having `max_size`... :p Ok, we have `max_size`. Then, may we want to set the default value of max_size to `127g` for rhel-7 and `0` for the other versions?

What makes me wonder most is `127g` is set in the test scripts in this pr. Is that the knowledge that customers should have for rhel-7? Maybe, it is...

richm:
> lgtm
> 
> Out of curiosity... Instead of introducing `max_size` and passing it by setting `127g` to `find_unused_disk.py`, we could quietly assign `127g` to an internal variable for rhel-7, I'd think.

The problem is that tests/provision.fmf lists 1TB disk sizes for the scsi (and nvme) disks, because we need those large sizes for some other tests.  So the tests that get an unused disk for swap need to somehow tell find_unused_disk.py to return an unused disk of less than a certain size:
```yaml
standard-inventory-qcow2:
  qemu:
    m: 2048
    drive:
      - size: 10737418240
      - size: 10737418240
      - size: 10737418240
      - size: 1099511627800
        interface: scsi
      - size: 1099511627800
        interface: scsi
      - size: 10737418240
        interface: scsi
      - size: 1099511627800
        interface: nvme
      - size: 10737418240
        interface: nvme
      - size: 10737418240
        interface: nvme
```
I'm not sure how to tell find_unused_disk.py to return an item from this list which has a scsi interface which is less than 128GB unless I introduce a `max_size` parameter.

> Most likely, I haven't understood the advantage of having `max_size`... :p Ok, we have `max_size`. Then, may we want to set the default value of max_size to `127g` for rhel-7 and `0` for the other versions?

No, because find_unused_disk.py is used for all of the tests, not just swap related tests.  For non-swap uses, there is no 128GB size restriction - only for the swap case.

> 
> What makes me wonder most is `127g` is set in the test scripts in this pr. Is that the knowledge that customers should have for rhel-7? Maybe, it is...

They should know that they cannot create a swap partition of size greater than 127GB.

nhosoi:
Thank you for the detailed explanations, @richm!

nhosoi (approved):
lgtm

vojtechtrefny (approved):
Looks good to me, thank you.

PR #340, opened 2023-03-14
Fix ansible-lint issues in tests

richm:
All tasks must be named
Task names must begin with uppercase letter
Fix Jinja spacing issues
Fix long lines with appropriate wrapping
Do not use `{{ .... }}` in `when` and `that` conditionals
Some tests were using `safe_mode` instead of `storage_safe_mode`
LUKS tests were using `regex_search` incorrectly and always
passing - use `is search` instead and fix pattern to parse
luksdump output correctly.
Refactor tests that check for the role to raise an error - add
a new file verify-role-failed.yml which checks for ansible errors
and blivet errors


richm:
[citest]

richm:
[citest]

PR #352, opened 2023-05-15
feat: User-specified mount point owner and permissions

japokorn:
Added new volume options related to mount point directory: mount_user - directory owner (string)
mount_group - directory group (string)
mount_permissions - directory permissions (string; same format as chmod)

Resolves: rhbz#2181661

PR #358, opened 2023-05-23
fix: add back systemd refresh after remove obsolete mounts

richm:
Turns out that the systemd refresh removed as part of
https://github.com/linux-system-roles/storage/pull/352
was needed.  This fix adds it back.


richm:
@japokorn @vojtechtrefny this is why tests_luks and tests_misc are currently broken.  After this merges, you'll need to rebase.

PR #359, opened 2023-05-24
test: use SYSTEM_ROLES_REMOVE_CLOUD_INIT=1 with remove-cloud-init

richm:
When testing with qemu and `--remove-cloud-init`, some of the tests
will fail because of missing packages.  However, if you define
`SYSTEM_ROLES_REMOVE_CLOUD_INIT=1` environment variable, you will
let the storage role determine which packages should be installed
and install them, which is the default.


richm:
[citest]

PR #396, opened 2023-10-09
fix: Do not remove swap at every run

scaronni:
Enhancement: The role is not idempotent if the system has swap. At every execution, the swap entries are remove and re-added in the next task.

Reason: There is specific code that deletes swap unconditionally. I could not find any apparent real reason for this piece of code to exist.

Result: Role is idempotent.

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -1647,8 +1647,6 @@ def get_mount_info(pools, volumes, actions, fstab):
         # handle removal of existing mounts of this volume
         if mount and mount['fs_type'] != 'swap' and mount['mount_point'] != volume['mount_point']:
             replace = dict(path=mount['mount_point'], state="absent")
-        elif mount and mount['fs_type'] == 'swap':
-            replace = dict(src=mount['device_id'], fstype="swap", path="none", state="absent")
 
         return mounted, replace
 


richm:
[citest]

richm:
well, the swap tests pass on el8 and el9 - as far as correctness I'll have to defer to @vojtechtrefny and @japokorn

scaronni:
Just to be clear, the blivet module removes the swap on the live system and then re-adds it back in the subsequent one, which means if a system is using swap heavily there might be some serious issues.

The swap is removed anyway correctly if people specify it as absent in the variables.

vojtechtrefny:
The special condition for swap was added in #90 @dwlehman do you remember why? Looking at the code I don't see a reason for always removing the swap entry.

dwlehman:
Sadly I do not have any record or memory of why to do that, but I think that if all the tests are passing this change is probably good.

vojtechtrefny (approved):

vojtechtrefny:
> Sadly I do not have any record or memory of why to do that, but I think that if all the tests are passing this change is probably good.

Ok, I think we can merge this. The tests are passing and we definitely don't want to always remove and then re-add the swap entry and I don't see another reason for having this condition.

PR #400, opened 2023-10-27
feat: Support for creating volumes without a FS

japokorn:
Currently whenever volume is created without fs_type specification, a filesystem of default type (i.e. xfs) is automatically put on it.
This change allows user to prevent FS creation by using "unformatted" value as a fs_type option. In the same manner it also allows to remove an existing FS. To be able to do that the safe mode has to be off.

Enhancement:

Reason:

Result:

Issue Tracker Tickets (Jira or BZ if any):

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,9 @@ variables:
   This indicates the desired file system type to use, e.g.: "xfs", "ext4", "swap".
   The default is determined according to the OS and release
   (currently `xfs` for all the supported systems).
+  Use "unformatted" if you do not want file system to be present.
+  __WARNING__: Using "unformatted" file system type on an existing filesystem
+               is a destructive operation and will destroy all data on the volume.
 
 - `fs_label`
 
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -20,7 +20,7 @@ version_added: "2.5"
 description:
     - "WARNING: Do not use this module directly! It is only for role internal use."
     - "Module configures storage pools and volumes to match the state specified
-       in input parameters. It does not do any management of /etc/fstab entries."
+       in input parameters."
 
 options:
     pools:
@@ -435,6 +435,11 @@ class BlivetVolume(BlivetBase):
 
     def _get_format(self):
         """ Return a blivet.formats.DeviceFormat instance for this volume. """
+
+        if self._volume['fs_type'].lower() == 'unformatted':
+            # Do not create any fs when user explicitly said so
+            return None
+
         fmt = get_format(self._volume['fs_type'],
                          mountpoint=self._volume.get('mount_point'),
                          label=self._volume['fs_label'],
@@ -560,7 +565,8 @@ class BlivetVolume(BlivetBase):
         """ Schedule actions as needed to ensure the volume is formatted as specified. """
         fmt = self._get_format()
 
-        if self._device.format.type == fmt.type:
+        if ((fmt is None and self._device.format.type is None)
+                or (fmt is not None and self._device.format.type == fmt.type)):
             # format is the same, no need to run reformatting
             dev_label = '' if self._device.format.label is None else self._device.format.label
             if dev_label != fmt.label:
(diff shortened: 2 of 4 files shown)


tests/tests_change_fs.yml
```diff
@@ -101,6 +101,22 @@
...
+      include_role:
+        name: linux-system-roles.storage
+      vars:
+        storage_pools:
+          - name: foo
+            disks: "{{ unused_disks }}"
+            volumes:
+              - name: test1
+                size: "{{ volume_size }}"
+                fs_type: None
```
richm, line 114: Does this require `storage_safe_mode: false`?  Seems kinda dangerous

richm:
What happens now if the fs_type is not specified? e.g. if you use
```yaml
        storage_pools:
          - name: foo
            disks: "{{ unused_disks }}"
            volumes:
              - name: test1
                size: "{{ volume_size }}"
```
what happens?  Does the role do nothing?  Does it remove the fs?  Does it create a volume with no fs? 

What happens when a user has a playbook with the above, which was working and creating file systems with fs_type `xfs`, and they upgrade the storage role, and run the playbook again?  The user expects that the behavior will not change - that is, it should create fs_type `xfs`.  We cannot easily break that, especially if the default will be destructive, or will just silently fail.

I think a better option to avoid breaking the api is to use a special value for fs_type e.g. `fs_type: NOFS` which will create the volume with no fs, or remove the fs on an existing volume.

japokorn:
Using None is in this case recognized and processed differently than when `fs_type` is not used at all.

For idempotency reasons we already have to be able to identify the situations when some variable is omitted. When that happens, any value found (on possibly already existing device) should remain unchanged.

Otherwise, running the role over pre-existing encrypted volume without for example `encryption` option in the task would force default `absent` value on it and change the device.

tests/tests_change_fs.yml
japokorn: Yes, it does. In code the behavior extends the functionality of `_reformat` method which handles the `safe_mode`.
Still, `fs_type: None` 'creation' on empty device will work even with `safe_mode` on since it is a no-op.

richm:
The way the current role (current contents of `main`, not this PR) works is this: if fs_type is omitted, or has the value of `null`, an fs is created with a type of `xfs` (on RHEL 8.9 manage nodes anyway).  *I don't think we should change this behavior* as it would break the existing API.  I have verified this by using the test `tests_create_disk_then_remove.yml` and using various different values for, or omitting, `fs_type`.

vojtechtrefny:
> I think a better option to avoid breaking the api is to use a special value for fs_type e.g. `fs_type: NOFS` which will create the volume with no fs, or remove the fs on an existing volume.

We are currently using `unformatted` in blivet-gui and `empty` in udisks so these would be also options. There is a (old and probably never really used) filesystem called NoFS so we should probably avoid using that.

japokorn:
Using ansible 'false' equivalents might be interchangeable with None, so based on discussion with vtrefny I have changed the string from None to all case variations of `unformatted`.

library/blivet.py
```diff
@@ -435,6 +435,11 @@ def _apply_defaults(self):
 
     def _get_format(self):
         """ Return a blivet.formats.DeviceFormat instance for this volume. """
+
+        if self._volume['fs_type'].lower() == 'unconfigured':
```
richm, line 439:
```suggestion
        if self._volume['fs_type'].lower() == 'unformatted':
```
?

README.md
```diff
@@ -143,6 +143,7 @@ variables:
   This indicates the desired file system type to use, e.g.: "xfs", "ext4", "swap".
   The default is determined according to the OS and release
   (currently `xfs` for all the supported systems).
+  Use "unformatted" if you do not want file system to be present.
```
richm, line 146: or maybe this should be `"unconfigured"`?

richm: Not sure how we note destructive operations elsewhere, but seems to me like we should say something like `**WARNING**: Using "unconfigured" on an existing filesystem will remove the filesystem.`

japokorn: Thanks. The correct value is actually 'unformatted' in the code. I made sure the value is unified.
I have also added a warning note into the README.

richm:
ok - I also did a test with

```yaml
        storage_pools:
          - name: foo
            disks: "{{ unused_disks }}"
            volumes:
              - name: test1
                size: "{{ volume_size }}"
                fs_type: unformatted
```
i.e. create a volume without an FS - it worked fine

richm (approved):

PR #409, opened 2023-12-14
Update enable_coprs.yml

real-joshua:
Add missing "when" statement

Enhancement:

This adds a "when" statement that is missing.

Reason:

The 1st and 3rd task check whether the variable is defined, but the 2nd does not. Therefore if the variable is not defined this task file's execution will fail at the 2nd task and the check that is part of the 3rd task will never have any effect.

Result:

Execution will no longer fail at the 2nd task if the variable is not defined, instead it will continue to the third which then again checks whether the variable is defined.

Issue Tracker Tickets (Jira or BZ if any):

### DIFF
--- a/tasks/enable_coprs.yml
+++ b/tasks/enable_coprs.yml
@@ -17,7 +17,7 @@
     use: "{{ (__storage_is_ostree | d(false)) |
              ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
   when: 
-    - _storage_copr_packages is defined
+    - _storage_copr_support_packages is defined
     - install_copr | d(false) | bool
 
 - name: Enable COPRs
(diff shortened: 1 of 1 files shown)


richm:
@real-joshua thanks - how are you seeing this error?  It looks like a real problem, but I cannot figure out how to reproduce it.
* What is your controller platform and ansible version?
* What is your managed node platform and version?

real-joshua:
Never mind, now that I read it again this makes no sense. I misread `_storage_copr_support_packages`  as `_storage_copr_packages`. :D

richm:
Did you mean to close this PR? I often hit the Close button when I mean to add a Comment.
If you didn't mean to close - my questions from https://github.com/linux-system-roles/storage/pull/409#issuecomment-1857119445 still apply

real-joshua:
> Did you mean to close this PR?

Yes, I noticed that my thought made no sense.

PR #410, opened 2023-12-15
ci: Use supported ansible-lint action; run ansible-lint against the collection

richm:
The old ansible-community ansible-lint is deprecated.  There is a
new ansible-lint github action.  The new ansible-lint has several checks
related to ansible-test and the ignore files.  Many of our ignore settings
are not allowed any more and are required to be fixed or addressed in the
Ansible preferred way.

The python imports have to be wrapped in a try/except ImportError, and
where possible, an error must be returned from the module explaining
what was not able to be imported.

The module documentation must comply with the Ansible standards.  One
aspect of this is the `version_added` must be a valid ansible-core
version in X.Y.Z format.  Note that this version isn't really used
anywhere, so it doesn't matter for users of the role, it is purely
an `ansible-test` and import gating issue.

I created a function `generate_module_doc` which will generate the
DOCUMENTATION options section from the Ansible module_args, if
necessary.  Hopefully it won't be necessary going forward.

I fought a losing battle with flake8, pep8, etc. in ansible-test
to try to preserve the formatting in size.py - future developers
are welcome to try.

The result of this is that the .sanity files can be reduced to the
bare minimum which will greatly reduce the maintenance burden of
those files, make it easier to support newer versions of Ansible,
and make it easier to import the system roles collection into Galaxy
and Automation Hub.

The latest Ansible repo gating tests run ansible-lint against
the collection format instead of against individual roles.
We have to convert the role to collection format before running
ansible-test.

Role developers can run this locally using
`tox -e collection,ansible-lint-collection`
See https://github.com/linux-system-roles/tox-lsr/pull/125

Add `---` doc start to .markdownlint.yaml

Fixed some other linter errors in some tests files.


richm:
[citest]

PR #425, opened 2024-02-09
feat: Added support for blivet fstab handling

japokorn:
Blivet can now handle fstab, so handling it in the role became
redundant. This adapts the role for the blivet change and removes no longer needed code.   
- requires blivet PRs [#1119](https://github.com/storaged-project/blivet/pull/1119) (main change) and [#1197](https://github.com/storaged-project/blivet/pull/1197) (bug fixes)
    
Issue Tracker Tickets: [#20182](https://issues.redhat.com/projects/RHEL/issues/RHEL-20182?filter=myopenissues)


tests/test-verify-volume-mount.yml
```diff
@@ -55,6 +55,8 @@
     msg: >-
       Found unexpected mount state for volume
       '{{ storage_test_volume.name }}' device
+      (expected matches '{{ storage_test_mount_expected_match_count|int }}';
```
richm, line 58:
```suggestion
      (expected matches '{{ storage_test_mount_expected_match_count | int }}';
```

tests/test-verify-volume-mount.yml
```diff
@@ -55,6 +55,8 @@
...
+      found matches '{{ storage_test_mount_device_matches|length }}')
```
richm, line 59:
```suggestion
      found matches '{{ storage_test_mount_device_matches | length }}')
```
ansible-lint likes spaces around the ` | ` expressions

richm:
Will the role still work on systems that do not have the right version of blivet?  That is, for managing older rhel7, rhel8, and rhel9 systems, will the role still need the "old" fstab handling?

japokorn:
> Will the role still work on systems that do not have the right version of blivet? That is, for managing older rhel7, rhel8, and rhel9 systems, will the role still need the "old" fstab handling?

No. Both blivet and storage role have to be updated for them to work.

tests/test-verify-volume-mount.yml
japokorn: Fixed.

tests/test-verify-volume-mount.yml
japokorn: Fixed.

japokorn:
> > > Will the role still work on systems that do not have the right version of blivet? That is, for managing older rhel7, rhel8, and rhel9 systems, will the role still need the "old" fstab handling?
> > 
> > 
> > No. Both blivet and storage role have to be updated for them to work.
> 
> We can't break the storage role for rhel7 and rhel8 users. If we want to use the new blivet fstab feature, we will have to change the role to support it, _and_ leave the old fstab handling in the blivet module.

[The latest changes](https://github.com/storaged-project/blivet/pull/1197) in blivet fstab handling should allow to keep the storage role as is.

In these circumstances, adapting SR to handle new blivet fstab and also making it working with the older version would mean creating redundancies in the role. The better way is just to let the role always handle the fstab the old way.

I'll give this a bit of grace period to make sure everything actually works. If it does I am going to close this PR.

PR #438, opened 2024-04-30
feat: PV resize support

japokorn:
There is an usecase when the physical device size can change (e.g. on
VM). We need to be able to change the size of the LVM PV to accomodate
that. This adds a new pool parameter `grow_to_fill`. When set, pool PVs
will try to take all available space on their respective devices.
Defaults to `False`.
    
Requires blivet version that supports this feature.
For tests this is checked by using `does_library_support` script.

Storage role development often relies on blivet library and changes in
it. Whether or not is specific feature supported by blivet is usually determined
by blivet version. That is a cumbersome process, especially when the feature
has not yet been added into blivet and the version has to be guessed.
    
Added script `does_library_support` verifies existence of the feature by using python
introspection and asking for existence of specific item in the library
(e.g. `blivet.formats.lvmpv.LVMPhysicalVolume.grow_to_fill`).
The script is supposed to be used for the tests only.

Blivet PR: [#1229](https://github.com/storaged-project/blivet/pull/1229)
JIRA issue: [STORAGECFG-743](https://issues.redhat.com/browse/STORAGECFG-743)

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -60,6 +60,9 @@ options:
             encryption_tang_thumbprint:
                 description: encryption_tang_thumbprint
                 type: str
+            grow_to_fill:
+                description: grow_to_fill
+                type: bool
             name:
                 description: name
                 type: str
@@ -379,7 +382,7 @@ try:
     from blivet3.callbacks import callbacks
     from blivet3 import devicelibs
     from blivet3 import devices
-    from blivet3.deviceaction import ActionConfigureFormat, ActionAddMember, ActionRemoveMember
+    from blivet3.deviceaction import ActionConfigureFormat, ActionResizeFormat, ActionAddMember, ActionRemoveMember
     from blivet3.devicefactory import DEFAULT_THPOOL_RESERVE
     from blivet3.flags import flags as blivet_flags
     from blivet3.formats import fslib, get_format
@@ -395,7 +398,7 @@ except ImportError:
         from blivet.callbacks import callbacks
         from blivet import devicelibs
         from blivet import devices
-        from blivet.deviceaction import ActionConfigureFormat, ActionAddMember, ActionRemoveMember
+        from blivet.deviceaction import ActionConfigureFormat, ActionResizeFormat, ActionAddMember, ActionRemoveMember
         from blivet.devicefactory import DEFAULT_THPOOL_RESERVE
         from blivet.flags import flags as blivet_flags
         from blivet.formats import fslib, get_format
@@ -421,6 +424,7 @@ if BLIVET_PACKAGE:
     blivet_flags.allow_online_fs_resize = True
     blivet_flags.gfs2 = True
     set_up_logging()
+
     log = logging.getLogger(BLIVET_PACKAGE + ".ansible")
 
     # XXX add support for LVM RAID raid0 level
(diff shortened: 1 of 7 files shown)


spetrosi:
[citest]

tests/test-verify-pool-members.yml
```diff
@@ -70,6 +70,23 @@
     loop_var: pv
   when: storage_test_pool.type == 'lvm'
 
+- name: Check that blivet supports PV grow to fill
+  ansible.builtin.script: >-
+    scripts/does_library_support.py
+    blivet.formats.lvmpv.LVMPhysicalVolume.grow_to_fill
+  register: grow_supported
+  changed_when: false
```
richm, line 78: Using `script` with python scripts is tricky because the python interpreter path is different on every platform.  The `snapshot` role uses `script` and has quite a bit of logic to determine which python interpreter to use.

https://github.com/linux-system-roles/snapshot/blob/main/tasks/main.yml#L21
The default value for `__snapshot_python` is defined here: https://github.com/linux-system-roles/snapshot/blob/main/vars/main.yml#L6

override for el7: https://github.com/linux-system-roles/snapshot/blob/main/vars/RedHat_7.yml#L4 and https://github.com/linux-system-roles/snapshot/blob/main/vars/CentOS_7.yml#L4

override for el8: https://github.com/linux-system-roles/snapshot/blob/main/vars/CentOS_8.yml#L4 and https://github.com/linux-system-roles/snapshot/blob/main/vars/RedHat_8.yml#L4

You might ask - well, why doesn't `script` use the Ansible python interpreter on the remote system?  If you can figure out how to do that, good - I could not figure out how to do that, which is why we implemented the `snapshot` role the way we did.

tests/test-verify-pool-members.yml
```diff
@@ -70,6 +70,23 @@
...
+
+- name: Verify that PVs fill the whole devices when they should
+  include_tasks: verify-pool-member-pvsize.yml
+  loop: "{{ _storage_test_pool_pvs | default([], true) }}"
```
richm, line 82: You shouldn't need to use the form `default([], true)` here - just `default([])` should work - if `_storage_test_pool_pvs` is defined, it should be a `list` - if it is an empty list (which evaluates to `false` in a boolean context, which is what the `, true` is for in the `default` filter), then nothing will be done - the loop will not be executed.

tests/verify-pool-member-pvsize.yml
```diff
@@ -0,0 +1,23 @@
...
+    that: (dev_size.bytes - actual_pv_size.stdout | int) |
+      abs / actual_pv_size.stdout | int < 0.04
+    msg: >-
+      PV resize failure; size difference too big
+      (device size: {{ dev_size.bytes }})
+      (actual PV size: {{ actual_pv_size.stdout }})
+
+- name: Clean up test variables
+  set_fact:
+    actual_pv_size: null
```
richm, line 23:
```suggestion
    actual_pv_size: null
    dev_size: null
```

japokorn: Fixed.
(Added missing `dev_size: null`)

tests/test-verify-pool-members.yml
japokorn: Fixed.
(Removed unnecessary `true`)

tests/test-verify-pool-members.yml
japokorn: I did a bit of testing and it seems that the value we are looking for is in `ansible_python.executable`.
I have added
```
      args:
        executable: "{{ ansible_python.executable }}"
```
to see how it fares.

japokorn:
[citest]

library/blivet.py
```diff
@@ -51,6 +51,9 @@
             encryption_password:
                 description: encryption_password
                 type: str
+            grow_to_fill:
+                description: grow_to_fill
+                type: bool
```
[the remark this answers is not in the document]
japokorn: Fixed. Added missing description to README.

library/blivet.py
```diff
@@ -1747,6 +1751,18 @@ def _manage_members(self):
...
         remove_disks = [pv for pv in self._device.pvs if not any(d in pv.ancestors for d in self._disks)]
 
+        if self._pool['grow_to_fill']:
+            grow_pv_candidates = [pv for pv in self._device.pvs if pv not in remove_disks and pv not in add_disks]
+
+            for pv in grow_pv_candidates:
+                if abs(self._device.size - self._device.current_size) < 2 * self._device.pe_size:
+                    continue
+
+                pv.format.update_size_info()  # set pv to be resizable
```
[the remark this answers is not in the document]
japokorn: Fixed. Added condition that skips resizing and logs warning if PV cannot be resized.

richm:
[citest]

japokorn:
Issues seem unrelated to this change since they are currently happening in the main branch as well.

richm:
please rebase on top of the latest main branch to pick up test fixes

richm (approved):

richm:
@japokorn please rebase on top of latest `main` branch and then we can merge

PR #447, opened 2024-06-05
test: lsblk can return LOG_SEC or LOG-SEC

richm:
get_unused_disk is broken on some systems because `lsblk ... LOG-SEC` can
return `LOG_SEC` with an underscore instead of the requested
`LOG-SEC` with a dash.


richm:
[citest]

PR #452, opened 2024-06-27
fix: add support for EL10

richm:
According to the Ansible team, support for listing platforms in
role `meta/main.yml` files is being removed.
Instead, they recommend using `galaxy_tags`

https://github.com/ansible/ansible/blob/stable-2.17/changelogs/CHANGELOG-v2.17.rst
"Remove the galaxy_info field platforms from the role templates"
https://github.com/ansible/ansible/issues/82453

Many roles already have tags such as "rhel", "redhat", "centos", and "fedora".
I propose that we ensure all of the system roles have these tags.
Some of our roles support Suse, Debian, Ubuntu, and others.
We should add tags for those e.g. the ssh role already has tags for "debian" and "ubuntu".

In addition - for each version listed under `platforms.EL` - add a tag like `elN`.

Q: Why not use a delimiter between the platform and the version e.g. `el-10`?

This is not allowed by ansible-lint:

```
meta-no-tags: Tags must contain lowercase letters and digits only., invalid: 'el-10'
meta/main.yml:1
```

So we cannot use uppercase letters either.

Q: Why not use our own meta/main.yml field?

No other fields are allowed by ansible-lint:

```
syntax-check[specific]: 'myfield' is not a valid attribute for a RoleMetadata
```

Q: Why not use some other field?

There are no other applicable or suitable fields.

Q: What happens when we want to support versions like `N.M`?

Use the word "dot" instead of "." e.g. `el10dot3`.
Similarly - use "dash" instead of "-".

We do not need tags such as `fedoraall`.
The `fedora` tag implies that the role works on all supported versions of fedora.
Otherwise, use tags such as `fedora40` if the role only supports specific versions.

In addition - for roles that have different variable files for EL9, create
the corresponding EL10 files.

Change find_unused_disk.py so that it will never return a nvme disk unless it
is specifically asked for that type of disk.  The issue was that every test
was using an nvme disk.


vars/AlmaLinux_10.yml
```diff
@@ -0,0 +1,14 @@
...
+  - libblockdev-dm
+  - libblockdev-lvm
+  - libblockdev-mdraid
+  - libblockdev-swap
+  - xfsprogs
+  - stratisd
+  - stratis-cli
+# vdo not yet available on el10
+#  - vdo
+#  - kmod-kvdo
```
vojtechtrefny, line 14: This will be just the `vdo` package for 10, the kernel module will be builtin in the kernel (and called dm-vdo).

.ostree/packages-runtime-CentOS-10.txt
```diff
@@ -0,0 +1,3 @@
+kmod-kvdo
+python3-blivet
+vdo
```
vojtechtrefny, line 3: Shouldn't this be also commented out for now?

vojtechtrefny (approved):
Other than the vdo/kvdo packages comments, looks good to me.

.ostree/packages-runtime-CentOS-10.txt
richm: commenting stuff out of these files is not supported, so I just removed them

vars/AlmaLinux_10.yml
vojtechtrefny: Yes. I think something like `modprobe --dry-run kvdo || modprobe --dry-run dm-vdo` should work.

richm: ok - once we have vdo support in el10, I will submit another PR to enable tests

vojtechtrefny (approved):

PR #476, opened 2024-08-19
test: use pvcreate with a timeout and verbose logging

richm:
the pvcreate call gives us problems, and different problems on different
platforms.  Use `timeout 30s` to avoid hangs when it tries
to read from stdin.  Use verbose `-vvv` to figure out why it
hangs/fails.  Use `-y` to make it not ask questions and continue
anyway.


richm:
[citest]

PR #496, opened 2024-12-11
test: skip fingerprint if fstab is not written

richm:
Some tests do not cause the role to touch /etc/fstab, so skip the
fingerprint check.


richm:
[citest]

PR #511, opened 2025-02-07
ci: Check spelling with codespell

spetrosi:
* You can ignore words inline by adding a comment like `# codespell:ignore word`.
* You can ignore words by adding them to the `.codespell_ignores` file.
* You can ignore files and directories by adding them with `skip = ` to the `.codespellrc` file.


.pandoc_template.html5
```diff
@@ -25,7 +25,7 @@ Copyright (c) 2017 GitHub Inc.
 
 "GitHub Pandoc HTML5 Template" is Copyright (c) Tristano Ajmone, 2017-2020,
 released under the MIT License (MIT); it contains readaptations of substantial
-portions of the following third party softwares:
+portions of the following third party software:
```
vojtechtrefny, line 28: If this is a part of the copyright/license, we probably shouldn't change it. But I am not a licensing expert.

vojtechtrefny (approved):
Looks good to me.

.pandoc_template.html5
richm: I would prefer not to change it, but if we don't change it, how do we tell codespell to ignore it?  Note that in English, `softwares` is incorrect - `software` is both singular and plural.

vojtechtrefny: If I understand it correctly, this file is generated and we are not really expected to change it, right? So I think we can just skip it in `.codespellrc`.

PR #514, opened 2025-03-03
fix: there is no kmod-vdo on EL10 and later

richm:
Cause: The blivet module adds `kmod-vdo` as a required package.

Consequence: The package fails to be installed on EL10 and later because
it was removed.

Fix: Ensure `kmod-vdo` is not added as a required package on EL10.

Result: The storage role does not give an error when using VDO
on EL10 and later.


richm:
[citest]

richm:
[citest_bad]

spetrosi (approved):

vojtechtrefny (approved):

PR #518, opened 2025-04-08
ci: several changes related to new qemu test, ansible-lint, python versions, ubuntu versions

richm:
There is a new QEMU based test which uses the qemu/kvm capability of
github action runners.  This is the basis for new bootc/image mode tests
which we will be rolling out in the near future.

ansible-lint requires that the collection path is set so that the requirements
it installs are installed in the correct place.

There has been some general github action deprecation of python versions and
ubuntu versions that we have had to fix.

Remove `CONTRIBUTOR` from the list of users who can trigger citest.

For more information, see

* https://github.com/linux-system-roles/.github/pull/98
* https://github.com/linux-system-roles/.github/pull/94
* https://github.com/linux-system-roles/.github/pull/93
* https://github.com/linux-system-roles/.github/pull/92
* https://github.com/linux-system-roles/.github/pull/91

PR #519, opened 2025-04-16
tests: Fix more cleanups; ci: Attach batch.txt and batch.report on failures

martinpitt:
This should fix the `tests_remove_mount.yml` and `tests_existing_lvm_pool.yml` follow-up failures from https://github.com/linux-system-roles/storage/pull/518#issuecomment-2803840784 . It should also make `tests_lvm_pool_pv_grow.yml` fail "correctly", i.e. with the "Unexpected difference between PV size" error (which you get when running  this in isolation), instead of the "/dev/foo: already exists in filesystem" which is due to a dirty VM.

### DIFF
--- a/.github/workflows/qemu-kvm-integration-tests.yml
+++ b/.github/workflows/qemu-kvm-integration-tests.yml
@@ -114,7 +114,10 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: "logs-${{ matrix.scenario.image }}-${{ matrix.scenario.env }}"
-          path: tests/*.log
+          path: |
+            tests/*.log
+            batch.txt
+            batch.report
           retention-days: 30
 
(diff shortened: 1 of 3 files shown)


richm (approved):

martinpitt:
Nice, as planned -- [the run](https://github.com/linux-system-roles/storage/actions/runs/14492638475/job/40652472706?pr=519) is down to three failures, and [tests_lvm_pool_pv_grow failed correctly](https://github.com/linux-system-roles/storage/actions/runs/14492638475/job/40652472706?pr=519#step:11:6802)

martinpitt:
Yes! I'm just not sure about that codecov failure, is that "expected"?

richm:
> Yes! I'm just not sure about that codecov failure, is that "expected"?

Yes

PR #527, opened 2025-05-07
ci: bump tox-lsr to 3.8.0; rename qemu/kvm tests

richm:
This will make the qemu/kvm tests be tested in either
ascending or descending ASCII order.  This should give
us better test coverage of clean up scenarios which may
fail depending on the order of the previous tests.

Rename the qemu/kvm tests so that the statuses are shorter
and more intuitive.

Improve qemu/kvm test failure error reporting.

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Install tox, tox-lsr
         run: |
           set -euxo pipefail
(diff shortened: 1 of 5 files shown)


martinpitt (changes requested):
The [resize failure](https://github.com/linux-system-roles/storage/actions/runs/14889264984/job/41817013029?pr=527) is known, not a regression.

martinpitt (approved):
... and this time the right button.

PR #546, opened 2025-07-21
fix: the encryption_key parameter should not be marked as no_log

richm:
Cause: The encryption_key parameter to the blivet module was marked as
no_log.  This causes the output value from the module to be replaced with the
string `VALUE_SPECIFIED_IN_NO_LOG_PARAMETER` which is then used as the value
for the key file in crypttab and in tests.

Consequence: Disk encryption using a key file does not work as the specified
key file is not used.

Fix: The encryption_key parameter does not need to be marked no_log because
it is just the name of the file, not the actual key contents.

Result: Disk encryption works, and the actual key file is used in crypttab
and elsewhere.

## Summary by Sourcery

Bug Fixes:
- Unmasked the encryption_key parameter to prevent it from being replaced with a placeholder value and ensure the real key file is used in crypttab and tests

### DIFF
--- a/library/blivet.py
+++ b/library/blivet.py
@@ -40,7 +40,7 @@ options:
                 description: encryption_cipher
                 type: str
             encryption_key:
-                description: encryption_key
+                description: encryption_key file, not the actual key
                 type: str
             encryption_key_size:
                 description: encryption_key_size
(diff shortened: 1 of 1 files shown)


spetrosi (approved):

spetrosi:
[citest]

richm:
[citest]

richm:
[citest]

PR #547, opened 2025-07-23
test: test luks encryption key

richm:
The storage_test_volume.encryption_key comes from the same variable that
is used to write the key to the crypttab - so change the test to test
against the original value passed in to the role.

## Summary by Sourcery

Improve LUKS encryption tests to verify against the original key and expand coverage across disk interfaces

Tests:
- Update encryption assertion to use __test_encryption_key variable for crypttab verification
- Add auto-generated playbooks for NVMe and SCSI variants of mixed and mixed_2 LUKS tests


richm:
This is the test for https://github.com/linux-system-roles/storage/pull/546

PR #575, opened 2025-12-31
fix: check for no disks specified and report correct error

richm:
Cause: The code did not check for missing or empty disks in the right place.

Consequence: The role would issue a strange error which did not indicate that the
problem is that no disks were specified.

Fix: Check for missing disks early and report appropriate error.

Result: The role will report an appropriate error message if no disks
are specified.

## Summary by Sourcery

Improve disk volume handling when no disks are specified and extend coverage with new storage tests.

Bug Fixes:
- Prevent disk volume processing from assuming at least one disk and return no device when the disks list is missing or empty.

Tests:
- Add regression test to verify the role fails with a clear error when no disks are specified for a volume.
- Extend default test configuration to include a volume with an empty disks list.
- Add generated NVMe and SCSI variants of multiple-partition DOS and GPT test playbooks.


richm:
[citest]

vojtechtrefny (dismissed):

richm:
testing farm ci is broken :-( investigating

richm:
[citest]

PR #577, opened 2026-01-05
refactor: handle INJECT_FACTS_AS_VARS=false by using ansible_facts instead

richm:
Ansible 2.20 has deprecated the use of Ansible facts as variables.  For
example, `ansible_distribution` is now deprecated in favor of
`ansible_facts["distribution"]`.  This is due to making the default
setting `INJECT_FACTS_AS_VARS=false`.  For now, this will create WARNING
messages, but in Ansible 2.24 it will be an error.

See https://docs.ansible.com/projects/ansible/latest/porting_guides/porting_guide_core_2.20.html#inject-facts-as-vars

## Summary by Sourcery

Update role variables, tasks, and tests to use the ansible_facts dictionary instead of legacy fact variables for compatibility with newer Ansible versions.

Enhancements:
- Replace direct fact variables (e.g., ansible_distribution, ansible_pkg_mgr, ansible_lvm, ansible_python, ansible_devices, ansible_architecture) with ansible_facts lookups across role vars and tasks.
- Extend required facts to include the package manager fact to support the new access pattern.

Tests:
- Update test playbooks and helper vars to reference facts via ansible_facts, aligning tests with the new fact access pattern and avoiding deprecation warnings.

### DIFF
--- a/tasks/enable_copr.yml
+++ b/tasks/enable_copr.yml
@@ -2,13 +2,13 @@
 - name: Get list of COPRs
   shell: |
     set -euo pipefail
-    {{ ansible_pkg_mgr }} repolist | \
+    {{ ansible_facts['pkg_mgr'] }} repolist | \
     grep -c `echo {{ repo.repository }} | tr / :` || true
   register: copr_present
   changed_when: false
 
 - name: Enable COPRs
   command:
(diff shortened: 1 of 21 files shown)


richm:
[citest]

vojtechtrefny (approved):

PR #583, opened 2026-01-30
fix: Volume with no size is same as size 100% - fixes divide by zero error

richm:
Cause: The code was treating a volume with no size as having size `0`.

Consequence: Some of the size calculations would use this as the denominator when
calculating percentage used or available, leading to a divide by zero error.

Fix: If no size is specified, treat this as if the size were specified as 100%.  That
is, use all of the available space in the pool.

Result: The code works as expected when no size is specified.

## Summary by Sourcery

Handle logical volumes without an explicit size by defaulting them to use all available space in the pool and document this behavior.

Bug Fixes:
- Prevent divide-by-zero errors when calculating usage for volumes that previously had an effective size of zero.

Documentation:
- Document that omitting a volume size is equivalent to specifying `size: 100%`.

Tests:
- Extend LVM percent size tests to cover volumes created without an explicit size and their removal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,9 @@ variables:
   When using `compression` or `deduplication`, `size` can be set higher than actual available space,
   e.g.: 3 times the size of the volume, based on duplicity and/or compressibility of stored data.
 
+  If `size` is not specified, it will be the same as if you specified `size: 100%` - it will use
+  all available space in the pool.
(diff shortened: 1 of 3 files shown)


vojtechtrefny (approved):
Looks good to me.

PR #593, opened 2026-02-28
test: ensure role gathers the facts it uses by having test clear_facts before include_role

richm:
The role gathers the facts it uses.  For example, if the user uses
`ANSIBLE_GATHERING=explicit`, the role uses the `setup` module with the
facts and subsets it requires.

This change allows us to test this.  Before every role invocation, the test
will use `meta: clear_facts` so that the role starts with no facts.  

Create a task file tests/tasks/run_role_with_clear_facts.yml to do the tasks
to clear the facts and run the role.  Note that this means we don't need to
use `gather_facts` for the tests.

Some vars defined using `ansible_facts` have been changed to be defined with
`set_fact` instead.  This is because of the fact that `vars` are lazily
evaluated - the var might be referenced when the facts have been cleared, and
will issue an error like `ansible_facts["distribution"] is undefined`.  This is
typically done for blocks that have a `when` condition that uses `ansible_facts`
and the block has a role invocation using run_role_with_clear_facts.yml
These have been rewritten to define the `when` condition using `set_fact`.  This
is because the `when` condition is evaluated every time a task is invoked in the
block, and if the facts are cleared, this will raise an undefined variable error.

## Summary by Sourcery

Ensure storage role tests run with cleared Ansible facts so the role gathers required facts itself before execution.

Enhancements:
- Make cryptsetup service masking conditional on service facts being present, gathering them only when needed.
- Extend the role’s required fact subsets to explicitly include device facts.

Tests:
- Run the storage role in all tests via a helper task file that clears facts before including the role, replacing direct include_role usage.
- Adjust test playbooks to no longer rely on service_facts-based skip checks and to set OS/version flags and helper vars using set_fact where needed.
- Update default test play and dependency test to invoke the role through the new helper, preserving public variable behavior.

### DIFF
--- a/tasks/main-blivet.yml
+++ b/tasks/main-blivet.yml
@@ -55,15 +55,10 @@
     extra_pkgs:
       - kpartx
 
-- name: Get service facts
-  service_facts:
-  when: storage_skip_checks is not defined or
-        not "service_facts" in storage_skip_checks
-
-# rejectattr required because the fix to service_facts is on Ansible > 2.12 only
-# https://github.com/ansible/ansible/pull/75326
-- name: Set storage_cryptsetup_services
-  set_fact:
+- name: Manage storage devices and check for errors
+  vars:
+    # rejectattr required because the fix to service_facts is on Ansible > 2.12 only
+    # https://github.com/ansible/ansible/pull/75326
     storage_cryptsetup_services: "{{
       ansible_facts.services.values() |
       selectattr('name', 'defined') |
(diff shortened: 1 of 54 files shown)


richm:
[citest]

tasks/main-blivet.yml
```diff
@@ -57,8 +57,7 @@
 
 - name: Get service facts
   service_facts:
-  when: storage_skip_checks is not defined or
-        not "service_facts" in storage_skip_checks
+  when: not "services" in ansible_facts
```
spetrosi: Why do you change the condition here and remove `service_facts` lines in playbooks?

richm: The role needs `service_facts` every time - the next task https://github.com/richm/linux-system-roles-storage/blob/457b0ca685d43692c52c0b5a2387295c24e293b7/tasks/main-blivet.yml#L64 will fail if `ansible_facts["services"]` is not defined - which worked the way the role and tests previously were written, and could assume that facts defined in the first role invocation would stay defined for the duration of the test, but is a hard fail if facts are cleared.  The `when` here should check if services is already defined i.e. the usual case when fact gathering is on.

However, looking at this some more

* the code should define `storage_cryptsetup_services` as a block var instead of using `set_fact` to avoid polluting the global namespace
* perhaps there is some optimization for tests that do not need to mask/unmask the cryptsetup services that could use something like the `storage_skip_checks` - maybe use `cryptsetup` instead of `service_facts` - I will investigate, but I think that may be an optimization that can wait for another PR

PR #614, opened 2026-04-16
feat: add role fingerprint to system journal

richm:
Feature: Add role fingerprint to system journal as the first task in the role.

Reason: The journal is often exported to reporting systems making it easy to
tell if the role is being used, and when.

Result: Users can get information in their reporting and log aggregation systems
of role usage.

## Summary by Sourcery

Add a custom fingerprinting mechanism that logs role begin and success markers to the system journal for the storage system role.

New Features:
- Introduce an sr_fingerprint Ansible module to write timestamped fingerprint messages to the system log.
- Record begin and success fingerprint messages for the storage role execution, including Ansible and OS version metadata.

Tests:
- Extend default role test playbook to capture start time and verify presence of begin and success fingerprints in the system journal.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -12,3 +12,9 @@
 
 - name: Include the appropriate provider tasks
   include_tasks: "main-{{ storage_provider }}.yml"
+
+- name: Record role success fingerprint
+  sr_fingerprint:
+    sr_message: >-
+      success system_role:storage ansible_version={{ ansible_version.full }}
+      {{ ansible_facts['distribution'] }}-{{ ansible_facts['distribution_version'] }}
--- a/.sanity-ansible-ignore-2.10.txt
+++ b/.sanity-ansible-ignore-2.10.txt
@@ -5,3 +5,4 @@ plugins/modules/find_unused_disk.py validate-modules:missing-gplv3-license
 plugins/modules/lvm_gensym.py validate-modules:missing-gplv3-license
 plugins/modules/resolve_blockdev.py validate-modules:missing-gplv3-license
 tests/storage/scripts/generate_tests.py shebang!skip
+plugins/modules/sr_fingerprint.py validate-modules:missing-gplv3-license
--- a/.sanity-ansible-ignore-2.11.txt
+++ b/.sanity-ansible-ignore-2.11.txt
@@ -5,3 +5,4 @@ plugins/modules/find_unused_disk.py validate-modules:missing-gplv3-license
 plugins/modules/lvm_gensym.py validate-modules:missing-gplv3-license
 plugins/modules/resolve_blockdev.py validate-modules:missing-gplv3-license
 tests/storage/scripts/generate_tests.py shebang!skip
+plugins/modules/sr_fingerprint.py validate-modules:missing-gplv3-license
--- a/.sanity-ansible-ignore-2.12.txt
+++ b/.sanity-ansible-ignore-2.12.txt
@@ -5,3 +5,4 @@ plugins/modules/find_unused_disk.py validate-modules:missing-gplv3-license
 plugins/modules/lvm_gensym.py validate-modules:missing-gplv3-license
 plugins/modules/resolve_blockdev.py validate-modules:missing-gplv3-license
 tests/storage/scripts/generate_tests.py shebang!skip
+plugins/modules/sr_fingerprint.py validate-modules:missing-gplv3-license
--- a/.sanity-ansible-ignore-2.13.txt
+++ b/.sanity-ansible-ignore-2.13.txt
@@ -5,3 +5,4 @@ plugins/modules/find_unused_disk.py validate-modules:missing-gplv3-license
 plugins/modules/lvm_gensym.py validate-modules:missing-gplv3-license
 plugins/modules/resolve_blockdev.py validate-modules:missing-gplv3-license
 tests/storage/scripts/generate_tests.py shebang!skip
+plugins/modules/sr_fingerprint.py validate-modules:missing-gplv3-license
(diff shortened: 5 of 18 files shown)


richm:
[citest]

richm:
[citest]

tasks/main.yml
```diff
@@ -1,4 +1,9 @@
 ---
+- name: Record storage role fingerprint in syslog
+  sr_fingerprint:
+    sr_message: "system_role:storage"
```
spetrosi: I am thinking about other useful info to be added with the fingerprint.
We can add timestamp in the message.
Also, maybe print Ansible version and platform for simplicity. Yes we can probably script getting this info from sos report content other than this line, but having it on the same line with the role fingerprint would make things easier, WDYT?
Would be interesting to see how many nodes ansible inventory had, I guess we can get this info on managed nodes by counting ansible_play_hosts.

richm: > I am thinking about other useful info to be added with the fingerprint. We can add timestamp in the message. 

This is what the message looks like in the journal:

```
Apr 17 10:38:55 hostname.domain.com python3.12[5169]: ansible-sr_fingerprint system_role:storage
```

I guess we could add an ISO 8601 format timestamp since the default doesn't have year or timezone:

```
Apr 17 10:38:55 hostname.domain.com python3.12[5169]:  ansible-sr_fingerprint system_role:storage 2026-04-17T08:42:40-06:00
```

> Also, maybe print Ansible version and platform for simplicity. Yes we can probably script getting this info from sos report content other than this line, but having it on the same line with the role fingerprint would make things easier, WDYT? Would be interesting to see how many nodes ansible inventory had, I guess we can get this info on managed nodes by counting ansible_play_hosts.

ok - I'll add ansible version and os-major.minor

richm: problem with adding platform: `{{ ansible_facts['distribution'] }}-{{ ansible_facts['distribution_version'] }}` - since this runs as the first task in the role, this runs *before* fact gathering - so these facts are not available - if we add the fingerprint after fact gathering, we don't get the benefit of having the fingerprint before everything else

richm: Here is the output with timestamp and ansible version:

```
Apr 17 11:33:12 hostname.domain.com python3.12[4732]: ansible-sr_fingerprint system_role:storage ansible_version=2.20.5rc1 2026-04-17T11:33:12-04:00
```

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest_bad]

tasks/main.yml
spetrosi: > problem with adding platform: `{{ ansible_facts['distribution'] }}-{{ ansible_facts['distribution_version'] }}` - since this runs as the first task in the role, this runs _before_ fact gathering - so these facts are not available - if we add the fingerprint after fact gathering, we don't get the benefit of having the fingerprint before everything else

To me, adding the fingerprint as the first thing is not really useful because it shows all attempts to run the role. If someone runs a role, fails, and never tries sysroles again, we still count them as our users.
I'd say let's add the fingerprint at the very end of a successful role run.

spetrosi: Why add fingerprint at the very beginning?

richm: > Why add fingerprint at the very beginning?

I thought it would be useful to know any and all role invocations at the very beginning, to see exactly when someone invoked the role.

Pros:

* know when someone invoked the role - can easily correlate this with other logs
* have a record of an attempt to invoke the role, even if the role fails or errors out

Cons:

* no facts available until after fact gathering - so not much available fingerprint metadata
* don't know if the role invocation was successful

I suppose we could combine the approaches - have a fingerprint at the beginning and at the end - the problem of doing it at the end is if you want to emit the fingerprint upon success or failure or error with the role result - yes, we could refactor the roles so that we run the role in a block/rescue/always - but there are some issues:

* if you use a `rescue`, you need to "re-raise" the error, but you cannot reconstruct the exact same `ansible_failed_result` to re-raise
* if you only use an `always`, I guess we could set a variable as the last task in the `block` to indicate that the role completed successfully, so we could report success or failure/error, but we couldn't report the exact error

spetrosi: Good point, well if we do two fingerprints for `init` and `success`, the absence of the `success` fingerprint indicates that the role failed. So no need to introduce rescue blocks everywhere.
Maybe still include the `init` fingerprint after collecting facts to provide platform?

richm: > Good point, well if we do two fingerprints for `init` and `success`, the absence of the `success` fingerprint indicates that the role failed. So no need to introduce rescue blocks everywhere. Maybe still include the `init` fingerprint after collecting facts to provide platform?

ok - I can add the init fingerprint just after fact gathering, and add the success fingerprint as the last task in tasks/main.yml

richm:
@spetrosi new implementation - here is what gets logged:

```
Apr 20 10:59:43 hostname.domain.com python3.12[5596]: ansible-sr_fingerprint fingerprint begin system_role:storage ansible_version=2.20.5rc1 CentOS-10 2026-04-20T10:59:43-04:00
Apr 20 11:00:02 hostname.domain.com python3.12[7666]: ansible-sr_fingerprint fingerprint success system_role:storage ansible_version=2.20.5rc1 CentOS-10 2026-04-20T11:00:02-04:00
```

spetrosi (dismissed):
The fingerprint looks very good!

richm:
[citest]

richm:
[citest]

PR #654, opened 2026-09-01
feat: Argument spec implementation for storage role

DonatSzabo:
Enhancement: Added argument spec and assert role spec validation to the storage role. Also wrote tests for it found in tests/tests_invalid_input.

Reason: Because it is a good addition to the linux-system-roles project.

Result: Successfully added it and prepared tests for it. I used AI during this implementation.

Issue Tracker Tickets (Jira or BZ if any): https://github.com/linux-system-roles/postfix/issues/206 https://redhat.atlassian.net/browse/RHELMISC-16008

## Summary by CodeRabbit

- **New Features**
  - Added documented configuration options and defaults for storage providers, volumes, pools, partitions, encryption, RAID, caching, and related settings.
  - Added supported-value restrictions for storage providers and RAID levels.
  - Added type checks for sizes, cache sizes, disk labels, and partition settings.

- **Bug Fixes**
  - Invalid or unsupported storage configurations are now rejected before changes begin.
  - Validation errors identify the affected setting and received value.

### DIFF
--- a/meta/argument_specs.yml
+++ b/meta/argument_specs.yml
@@ -10,7 +10,9 @@ argument_specs:
     options:
       storage_provider:
         type: str
-        default: "blivet"
+        default: blivet
+        choices:
+          - blivet
         description: >
           The storage management provider used by the role.
           Currently only `blivet` is supported.
@@ -25,8 +27,9 @@ argument_specs:
         type: raw
         default: null
         description: >
-          Whether to use partition-based layout. When `null`, the
-          role selects an appropriate layout automatically.
+          Whether to use partition-based layout. Accepts `true`,
+          `false`, or `null`. When `null`, the role selects an
+          appropriate layout automatically.
       storage_disklabel_type:
         type: raw
         default: null
@@ -151,12 +154,19 @@ argument_specs:
               and `path`.
           raid_level:
             type: str
+            choices:
+              - linear
+              - raid0
+              - raid1
+              - raid4
+              - raid5
+              - raid6
+              - raid10
+              - striped
+              - mirror
             description: >
-              RAID level for the volume. For `raid` type volumes,
-              accepted values are `linear`, `raid0`, `raid1`,
-              `raid4`, `raid5`, `raid6`, `raid10`. For LVM RAID,
-              accepted values also include `striped` and
-              `mirror`.
+              RAID level for the volume. Values `striped` and
+              `mirror` apply only to LVM RAID volumes.
           raid_device_count:
             type: int
             description: >
--- a/meta/argument_specs.yml
+++ b/meta/argument_specs.yml
@@ -1,0 +1,492 @@
+# SPDX-License-Identifier: MIT
+---
+argument_specs:
+  main:
+    short_description: The storage role.
+    description: >
+      The storage role allows you to configure local storage
+      including file systems, LVM volume groups, Stratis pools,
+      RAID arrays, and encrypted volumes with minimal input.
+    options:
+      storage_provider:
+        type: str
+        default: "blivet"
+        description: >
+          The storage management provider used by the role.
+          Currently only `blivet` is supported.
+      storage_safe_mode:
+        type: bool
+        default: true
+        description: >
+          Whether to operate in safe mode. When `true`, the role
+          fails instead of implicitly removing existing devices
+          or formatting.
+      storage_use_partitions:
+        type: raw
+        default: null
+        description: >
+          Whether to use partition-based layout. When `null`, the
+          role selects an appropriate layout automatically.
+      storage_disklabel_type:
+        type: raw
+        default: null
+        description: >
+          The disk label type to use for partitioning, such as
+          `gpt` or `msdos`. When `null`, the role selects an
+          appropriate label type automatically.
+      storage_udevadm_trigger:
+        type: bool
+        default: false
+        description: >
+          Whether to use udevadm trigger to cause udev changes
+          to take effect immediately. May help on some platforms
+          with udev issues.
+      storage_volumes:
+        type: list
+        elements: dict
+        default: []
+        description: >
+          List of standalone volumes to manage. Each volume
+          represents a storage device with a file system, such
+          as a disk, partition, or LVM logical volume.
+        options: &volume_options
+          name:
+            type: str
+            required: true
+            description: >
+              The name of the volume.
+          type:
+            type: str
+            choices:
+              - lvm
+              - disk
+              - partition
+              - raid
+            description: >
+              The type of volume. Valid values are `lvm` for LVM
+              logical volumes, `disk` for whole disk volumes,
+              `partition` for partition-based volumes, and `raid`
+              for RAID volumes.
+          state:
+            type: str
+            choices:
+              - present
+              - absent
+            description: >
+              Whether the volume should exist (`present`) or be
+              removed (`absent`).
+          size:
+            type: raw
+            description: >
+              The size of the volume. Accepts human-readable size
+              notation such as `10G` or `500M`, a percentage of
+              pool size such as `50%`, or `0` to use all
+              available space.
+          disks:
+            type: list
+            elements: str
+            description: >
+              List of disks to use as backing storage for the
+              volume. For `disk` type volumes, this must contain
+              exactly one item.
+          fs_type:
+            type: str
+            description: >
+              The desired file system type, such as `xfs`, `ext4`,
+              or `swap`. Use `unformatted` to leave the volume
+              without a file system.
+          fs_label:
+            type: str
+            description: >
+              Label for the file system.
+          fs_create_options:
+            type: str
+            description: >
+              Custom arguments to pass to `mkfs`.
+          fs_overwrite_existing:
+            type: bool
+            description: >
+              Whether to overwrite an existing file system if one
+              is detected on the volume.
+          mount_point:
+            type: str
+            description: >
+              Directory path where the file system will be
+              mounted.
+          mount_options:
+            type: str
+            description: >
+              Custom mount options as a string, such as `ro` or
+              `noatime`.
+          mount_user:
+            type: str
+            description: >
+              Desired owner of the mount directory.
+          mount_group:
+            type: str
+            description: >
+              Desired group of the mount directory.
+          mount_mode:
+            type: str
+            description: >
+              Desired permissions of the mount directory.
+          mount_check:
+            type: int
+            description: >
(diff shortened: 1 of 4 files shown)


meta/argument_specs.yml
```diff
@@ -0,0 +1,492 @@
...
+argument_specs:
+  main:
+    short_description: The storage role.
+    description: >
+      The storage role allows you to configure local storage
+      including file systems, LVM volume groups, Stratis pools,
+      RAID arrays, and encrypted volumes with minimal input.
+    options:
+      storage_provider:
+        type: str
```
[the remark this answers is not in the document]
richm: @DonatSzabo please add `choices` with `blivet` as the only choice

meta/argument_specs.yml
```diff
@@ -0,0 +1,492 @@
...
+        default: "blivet"
```
richm:
```suggestion
        default: blivet
```
In YAML you should not quote simple strings

meta/argument_specs.yml
```diff
@@ -0,0 +1,492 @@
...
+        default: true
+        description: >
+          Whether to operate in safe mode. When `true`, the role
+          fails instead of implicitly removing existing devices
+          or formatting.
+      storage_use_partitions:
+        type: raw
+        default: null
+        description: >
+          Whether to use partition-based layout. When `null`, the
```
richm: Add some sort of comment saying that this value is `true`, `false`, or `null`.

tasks/assert_role_vars.yml
```diff
@@ -0,0 +1,85 @@
...
+        or (storage_disklabel_type is string)
+    fail_msg: >-
+      storage_disklabel_type must be null or a string,
+      got {{ storage_disklabel_type | type_debug }}
+
+- name: Assert size in storage_volumes is a string or integer
+  ansible.builtin.assert:
+    that:
+      - >-
+        item.size is string
```
richm, line 28: if this is a string, it should be either `"0"` or match "^[1-9][0-9]*"

tasks/assert_role_vars.yml
```diff
@@ -0,0 +1,85 @@
...
+  loop_control:
+    index_var: idx
+    label: "{{ item.name | d('unnamed') }}"
+  when: item.size is defined
+
+- name: Assert cache_size in storage_volumes is a string or integer
+  ansible.builtin.assert:
+    that:
+      - >-
+        item.cache_size is string
```
richm, line 48: see above for `size`

tasks/assert_role_vars.yml
```diff
@@ -0,0 +1,85 @@
...
+  loop_control:
+    index_var: idx
+    label: "{{ item.name | d('unnamed') }}"
+  when: item.cache_size is defined
+
+- name: Assert size in pool volumes is a string or integer
+  ansible.builtin.assert:
+    that:
+      - >-
+        item.1.size is string
```
richm, line 69: see above for `size`

tasks/assert_role_vars.yml
```diff
@@ -0,0 +1,85 @@
...
+       | subelements('volumes', skip_missing=True) }}
+  loop_control:
+    label: "{{ item.0.name | d('unnamed') }}/{{ item.1.name | d('unnamed') }}"
+  when: item.1.size is defined
+
+- name: Assert cache_size in pool volumes is a string or integer
+  ansible.builtin.assert:
+    that:
+      - >-
+        item.1.cache_size is string
```
richm, line 90: see above for `size`

tests/tests_invalid_input.yml
```diff
@@ -0,0 +1,603 @@
...
+              block:
+                - name: Run role with non-bool storage_safe_mode
+                  ansible.builtin.include_role:
+                    name: linux-system-roles.storage
+                  vars:
+                    storage_safe_mode: "not_a_bool"
+              rescue:
+                - name: Mark invalid storage_safe_mode type rejected
+                  ansible.builtin.set_fact:
+                    __invalid_input_safe_mode_type_failed: true
```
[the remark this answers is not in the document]
richm: @DonatSzabo please add a check for the correct error message or pattern in `ansible_failed_result.msg`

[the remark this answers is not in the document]
richm: same with the other checks below

tasks/assert_role_vars.yml
richm: I think the condition should be something like this:

(item.size is string and (item.size == "0" or item.size is match(...))
or ((item.size | type_debug) == "int" and item.size >= 0)

You shouldn't need the `| int` because you have already determined that the value is type `int`

DonatSzabo: Thats true. I will get on that

richm:
@DonatSzabo you'll need to make these changes to make the tests pass:

```diff
diff --git a/tests/tests_resize.yml b/tests/tests_resize.yml
index 77d0aa6..4f49de4 100644
--- a/tests/tests_resize.yml
+++ b/tests/tests_resize.yml
@@ -10,7 +10,6 @@
     volume_size_after: '9g'
     invalid_size1: xyz GiB
     invalid_size2: none
-    unused_disk_subfact: "{{ ansible_facts['devices'][unused_disks[0]] }}"
     too_large_size: '{{ unused_disk_subfact.sectors | int * 1.2 * 512 }}'
     acc_large_size: '{{ unused_disk_subfact.sectors | int * 1.015 * 512 }}'
     acc_small_size: '{{ unused_disk_subfact.sectors | int * 0.985 * 512 }}'
@@ -25,6 +24,10 @@
       vars:
         __storage_get_package_facts: true
 
+    - name: Set unused_disk_subfact
+      set_fact:
+        unused_disk_subfact: "{{ ansible_facts['devices'][unused_disks[0]] }}"
+
     # For ext4 FS
 
     - name: Test ext4
@@ -140,7 +143,7 @@
         - name: Test for correct handling of invalid size specification
           include_tasks: verify-role-failed.yml
           vars:
-            __storage_failed_regex: invalid.+size
+            __storage_failed_exception: size must be a non-negative integer or a size string starting with a digit
             __storage_failed_msg: >-
               Unexpected behavior w/ invalid volume size
             __storage_failed_params:
@@ -156,7 +159,7 @@
         - name: Test for correct handling of invalid size specification - 2
           include_tasks: verify-role-failed.yml
           vars:
-            __storage_failed_regex: invalid.+size
+            __storage_failed_exception: size must be a non-negative integer or a size string starting with a digit
             __storage_failed_msg: >-
               Unexpected behavior w/ invalid volume size
             __storage_failed_params:
diff --git a/tests/verify-role-failed.yml b/tests/verify-role-failed.yml
index 9487834..f5fb32e 100644
--- a/tests/verify-role-failed.yml
+++ b/tests/verify-role-failed.yml
@@ -49,7 +49,8 @@
           msg_stdout is search(__storage_failed_exception) or
           msg_stderr is search(__storage_failed_exception) or
           stdout is search(__storage_failed_exception) or
-          stderr is search(__storage_failed_exception)
+          stderr is search(__storage_failed_exception) or
+          exception_msg_in_list
       when: __storage_failed_exception is defined
       vars:
         result_msg: "{{ ansible_failed_result.msg | d('') }}"
@@ -59,3 +60,5 @@
         msg_stderr: "{{ ansible_failed_result.msg.module_stderr | d('') }}"
         stdout: "{{ ansible_failed_result.module_stdout | d('') }}"
         stderr: "{{ ansible_failed_result.module_stderr | d('') }}"
+        exception_msg_in_list: "{{ ansible_failed_result.results | d([]) |
+          selectattr('msg', 'search', __storage_failed_exception) | list | length > 0 }}"
```

richm:
[citest_all]

richm:
[citest_bad]

meta/argument_specs.yml
```diff
@@ -0,0 +1,502 @@
...
+              - label
+              - path
+            description: >
+              Method used to identify the device in
+              `/etc/fstab`. Valid values are `uuid`, `label`,
+              and `path`.
+          raid_level:
+            type: str
+            choices:
+              - linear
```
richm, line 158: `"null"` - the string value, not the `null` value, is a supported value here - https://github.com/linux-system-roles/storage/blob/main/tests/tests_null_raid_pool.yml#L31

richm:
@DonatSzabo rewrite tests_missing_volume_type_in_pool.yml to use `verify-role-failed.yml` with the expected error message pattern in `__storage_failed_exception` instead of calling `tasks/run_role_with_clear_facts.yml` and checking for the blivet error message - with the new argument checking code, you will get an ansible error instead of a blivet error.

richm:
@DonatSzabo https://github.com/linux-system-roles/storage/pull/654#issuecomment-5670812933 same with tests_misc.yml - the error message is now coming from the argument validation instead of blivet

richm:
> @DonatSzabo [#654 (comment)](https://github.com/linux-system-roles/storage/pull/654#issuecomment-5670812933) same with tests_misc.yml - the error message is now coming from the argument validation instead of blivet

Same with tests_lvm_errors.yml

richm:
> > @DonatSzabo [#654 (comment)](https://github.com/linux-system-roles/storage/pull/654#issuecomment-5670812933) same with tests_misc.yml - the error message is now coming from the argument validation instead of blivet
> 
> Same with tests_lvm_errors.yml

And several more tests.

I know the storage role is kind of a pain because there are many, many tests, and the tests take a long time to run. But it will be faster to run the tests locally like this:

```bash
tox -e qemu-ansible-core-2-16 -- --image-name centos-9 --log-level debug --make-batch
```

Then try to fix as many errors in as many tests as you can before you run this again.  Note that when you run it again, you should first do `rm -rf batch.* tests/*.log`.

DonatSzabo:
[citest_all]

richm:
[citest_bad]

richm:
[citest_tft]

richm:
@DonatSzabo  the ansible 2.9 tests fail because argument_specs.yml is ignored in ansible 2.9 - therefore the tests that check if the role fails properly are failing because the error is being returned from blivet rather than the argument spec checking - for example, there is a test that tries to set `raid_level: fail` - the error returned is from blivet instead of argument_specs.yml checking or assert_role_vars.yml.

My suggestion is to change the test to look for a different error message when using ansible 2.9

The other way that I do not suggest is to add parameter checking for everything in assert_role_vars.yml - that is a lot of code to add for a platform that will go away soon

PR #660, opened 2026-09-04
fix: Move mount calls to a shared include task file

spetrosi:
Enhancement: Move the vendored and ansible.posix mount calls in main-blivet.yml into a shared include_mount.yml that is pulled in with include_tasks, passing module arguments as a single dict variable.
Reason: An ansible.posix FQCN in a statically parsed task file aborts play parsing on ansible 2.9 even when the task is skipped, which previously forced bare module names with an fqcn lint suppression.
Result: The collection module is now referenced by FQCN and only parsed at runtime, keeping ansible 2.9 support while dropping the bare-name workaround.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mount management across systems using Python 2 and Python 3.
  * Prevented compatibility issues on older systems where the modern mount module is unavailable.
  * Mount removal and setup now work more reliably across supported environments.

* **Refactor**
  * Consolidated mount handling into a shared workflow while preserving existing behavior.
  * Mount operations now select the appropriate implementation dynamically for each system.
  * Improved compatibility with older automation environments during mount operations.


tasks/main-blivet.yml
```diff
@@ -190,76 +190,46 @@
...
-  mount:  # noqa fqcn
-    src: "{{ mount_info['src'] | default(omit) }}"
-    path: "{{ mount_info['path'] }}"
-    fstype: "{{ mount_info['fstype'] | default(omit) }}"
-    opts: "{{ mount_info['opts'] | default(omit) }}"
-    state: "{{ mount_info['state'] }}"
+  include_tasks: include_mount.yml
+  vars:
+    __storage_mount_args:
+      src: "{{ mount_info['src'] | default(omit) }}"
```
richm, line 199: I think `omit` only works in the context of the actual module call - it passes a special per-module value to the module that the module knows to ignore - it does not tell Ansible/yaml/jinja to actually omit the key/value pair from the list of arguments.  I think the behavior is undefined, and possibly depends on which version of Ansible you are using, when there is no `src` in `mount_info`.

One way to do this cleanly is kind of ugly:

```yaml
  vars:
    # put the required arguments in the dict first, then conditionally add the
    # optional arguments
    __storage_mount_args: "{{ {'path': mount_info['path'], 'state': mount_info['state']} | combine(__maybe_src) | combine(__maybe_fstype) | combine(__maybe_opts) }}"
    __maybe_src: "{{ {'src': mount_info['src']} if mount_info.get('src', '') | length > 0 else {} }}"
    __maybe_fstype: "{{ {'fstype': mount_info['fstype']} if mount_info.get('fstype', '') | length > 0 else {} }}"
    __maybe_opts: "{{ {'opts': mount_info['opts']} if mount_info.get('opts', '') | length > 0 else {} }}"
```

richm: actually, in this case, using `omit` might just work - because you are using `omit` in the `vars` definition and because `vars` use lazy evaluation, the `omit` will not be evaluated until the module context

spetrosi:
[citest]

spetrosi:
[citest_all]

richm (approved):

